PR #3, opened 2016-04-12
Spotify

rosscreighton:


server/spotify.js
```diff
@@ -0,0 +1,168 @@
+export class SpotifyAPI {
```
rosscreighton: so one big thing I realized is that we don't need to use a class for this, and, since imports/exports are now available natively in javascript, its probably considered bad practice to do this. the es6 way to organize this code would be to just  put all of these functions/constants in a file, and export the ones you need. this way, we don't have to instantiate a class, and we don't have to refer to that class when calling a method we need. i felt like this was hard to explain so i converted spotify.js to the es6 and put it in this [pull request](https://github.com/greymatter0101/gm/pull/4)

it was fucking great that you had written those tests bc I could confirm the code was still working after the change.

server/spotify.js
```diff
@@ -0,0 +1,168 @@
...
+    this.baseUrl = 'https://api.spotify.com/v1'
+  };
+
+  search(type, searchItems) {
+
+    let items = [];
+    let ids = searchItems.map (si => si.id);
+    let response = this.get(`${this.baseUrl}/${type}?ids=${ids.join(',')}`);
+
+    for(let i = 0; i < response[type].length; i++) {
```
rosscreighton: probably want to use forEach here, will clean up the code a lot

server/spotify.js
```diff
@@ -0,0 +1,168 @@
...
+      items.push(item);
+    };
+
+    return items;
+
+  };
+
+  fuzzySearch(query) {
+
+    let q = `search?type=artist,album&q=${query}`;
```
rosscreighton: let should only be used when the variable will be reassigned

server/spotify.js
```diff
@@ -0,0 +1,168 @@
...
+    for(let i = 0; i < this.artistsFields.length; i++) {
+      let f = this.artistsFields[i];
+      artist[f] = f == 'name' ? json[f].replace('.', '') : json[f];
+    };
+
+    return artist;
+
+  };
+
+  translateAlbumsJson(json) {
```
rosscreighton: i'm not sure I understand whats going on with these translate__Json methods. can you explain?

server/spotify.tests.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('searches query should return results', function () {
+    const response = spotify.fuzzySearch('david bowie');
+    let boolean = response.albums.length > 0 && response.artists.length > 0;
+
+    chai.expect(boolean).to.equal(true);
+  });
+
+  it('album query should return results', function () {
+    let boolean = true;
```
rosscreighton: should this variable be named 'success'? looks like it can never get set to false.

server/spotify.tests.js
```diff
@@ -0,0 +1,43 @@
...
+    for(let i = 0; i < spotify.albumsFields.length; i++) {
+      let f = spotify.artistsFields[i];
+      if (typeof f == undefined) { success = false };
+    };
+
+    chai.expect(boolean).to.equal(true);
+  });
+
+  it('artist query should return results', function () {
+    let boolean = true;
```
rosscreighton: success?

server/spotify.tests.js
```diff
@@ -0,0 +1,43 @@
...
+
+    const davidBowieId = '0oSGxfWSnnOXhD2fKuz2Gy';
+    const response = spotify.artist(davidBowieId);
+
+    for(let i = 0; i < spotify.artistsFields.length; i++) {
+      let f = spotify.artistsFields[i];
+      if (typeof f == undefined) { success = false };
+    };
+
+    chai.expect(boolean).to.equal(true);
```
rosscreighton: this looks good to me but just a thought - there is probably a utility in chai that lets us make assertions on the key/value pairs (the shape) of the response object so we don't have to iterate over it.

server/spotify.tests.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('artist query should return results', function () {
+    let boolean = true;
+
+    const davidBowieId = '0oSGxfWSnnOXhD2fKuz2Gy';
+    const response = spotify.artist(davidBowieId);
+
+    for(let i = 0; i < spotify.artistsFields.length; i++) {
+      let f = spotify.artistsFields[i];
+      if (typeof f == undefined) { success = false };
```
rosscreighton: should this line be `if (typeof response[f] == undefined) { success = false }`? it looks like f will always be defined, whether the request returns results or not.

server/spotify.js
```diff
@@ -0,0 +1,168 @@
...
+    return {
+      albums: albums,
+      artists: artists
+    };
+
+  };
+
+  albumsSearch(query) {
+
+    let search_query = `${this.baseUrl}/search?type=album&q=${query}`;
```
rosscreighton: we should camelCase variable names

rosscreighton:
i like all of the method names and  i feel like everything is broken out in the right spots. i was able to understand whats going on really quickly

rosscreighton:
this spotify.js script probably belongs in `/imports/spotify/`

i don't think that directory exists on this branch lol but just to put a bug in your ear

PR #5, opened 2016-05-18
Artist publications

rosscreighton:


imports/startup/server/index.js
```diff
@@ -1 +1,2 @@
 import './fixtures.js';
+import './register-api.js';
```
sameckmeier: i feel like api.js would do, unless you think appending register- makes this file name clearer

imports/api/Artists/Artists.tests.js
```diff
@@ -1,30 +1,34 @@
...
-      const createArtist = (props = {}) => {
-        const artist = Factory.create('artist', props);
-      };
-
       before(function () {
-        Artists.remove({});
-        createArtist();
+        resetDatabase();
+
+        Factory.define('artist', Artists, {
```
sameckmeier: would be cool to put all factories into a module and export them as needed

rosscreighton:
i think it makes sense to close this one, we'll come back to it when we're ready for artist data

PR #7, opened 2016-06-19
page component

rosscreighton:
TL;DR: don't be confused by the branch name.

the name of this branch is confusing. i called it `application-layout` because i was originally thinking of this page component as a 'layout' because it renders a header and footer like a layout would. however i was thinking about it and this component will do more than provide standardized layout - it will handle user data, analytics, etc., so it is a component. 

alos, right now, just a `Page` component will serve us well, but in the future as we add more functionality to this component we may want to break it out into an `Application` component and `Page` component.

there are also some performance tweaks to `AlbumsContainer` in this PR, which cut the `AlbumPage` re-renders from 4 to 2. see commit messages for more detail.


imports/ui/containers/AlbumsContainer.jsx
```diff
@@ -23,8 +23,8 @@ export default AlbumsContainer = createContainer(({ params }) => {
   }
 
   return {
-    album: albumReady ? album : {},
```
sameckmeier: v nice

imports/ui/containers/AlbumsContainer.jsx
```diff
@@ -10,9 +10,9 @@ const spotifyAlbum = new ReactiveVar(null);
 export default AlbumsContainer = createContainer(({ params }) => {
   const albumsHandle = Meteor.subscribe('album', params.slug);
   const album = Albums.findOne({slug: params.slug});
+  const spotifyAlbumLoading = spotifyAlbum.get() === null;
```
sameckmeier: right right, nice

imports/ui/containers/AlbumsContainer.jsx
```diff
@@ -9,10 +9,12 @@ const spotifyAlbum = new ReactiveVar(null);
 
 export default AlbumsContainer = createContainer(({ params }) => {
   const albumsHandle = Meteor.subscribe('album', params.slug);
-  const album = Albums.findOne({slug: params.slug});
+  const albumReady = albumsHandle.ready();
```
sameckmeier: fantastic change. we should probs remove the second, embedded albumChange

imports/ui/pages/AlbumsPage.jsx
```diff
@@ -2,9 +2,12 @@ import React, { Component, PropTypes } from 'react';
 
 export default class AlbumsPage extends Component {
   render() {
+    const { album, spotifyAlbum } = this.props;
```
sameckmeier: i really like this. we should make a card with a checklist of some primary code quality todos, one of which is unpacking props/args

PR #8, opened 2016-06-19
run spotify tests on the client

rosscreighton:


imports/spotify/spotify.tests.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-env mocha */
```
sameckmeier: this is great, though, how have ya been running these on the client?

PR #9, opened 2016-06-19
creates a super basic database seeder.

rosscreighton:
it just creates documents when server starts based on the definitions in the `documentsToCreate` array. something this simple should work for us for quite some time, until we need to start seeding really complex data.


imports/lib/database.js
```diff
@@ -0,0 +1,23 @@
+import { Albums } from '../api/Albums/Albums';
```
sameckmeier: 👍

PR #13, opened 2016-07-11
Album page grid

sameckmeier:


.meteor/packages
```diff
@@ -33,4 +33,9 @@ aldeed:template-extension
 accounts-password
 react-meteor-data
 johanbrook:publication-collector
+
 static-html
+johnantoni:meteor-normalize
```
rosscreighton: What's meteor normalize?

.meteor/packages
```diff
@@ -10,7 +10,7 @@ mongo                   # The database Meteor supports right now
 reactive-var            # Reactive variable for tracker
 tracker                 # Meteor's client-side reactive programming library
 
-standard-minifier-css   # CSS minifier run for production mode
+seba:minifiers-autoprefixer
```
rosscreighton: Do we need this if we're using bourbon?

imports/lib/database.js
```diff
@@ -1,4 +1,4 @@
-import { Albums } from '../api/Albums/Albums';
+import { Albums } from '../api/albums/Albums';
```
rosscreighton: Are you current with master? I believe I uppercased all the collection directories on master but I could be wrong

.meteor/packages
sameckmeier: huh, weird. it looks like a css framework. definately shouldn't be in there.

imports/startup/client/routes.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Router, Route, browserHistory } from 'react-router';
-import Page from '../../ui/Page';
+import Page from '../../ui/components/Page/Page';
```
rosscreighton: Definitely makes sense to move Page but I wonder if it belongs in a layouts directly or something like that

imports/lib/database.js
sameckmeier: i dont believe so and wonderful change 👍

.meteor/packages
sameckmeier: we do. it's listed as package to include with bourbon.

imports/ui/components/AlbumsPage/AlbumsPage.jsx
```diff
@@ -0,0 +1,30 @@
+import React, { Component, PropTypes } from 'react';
+import './AlbumsPage.scss';
+
+export default class AlbumsPage extends Component {
+  render() {
+    const { album, spotifyAlbum } = this.props;
+
+    return(
+      <div className='album-page'>
```
rosscreighton: Looks good 👌🏻

rosscreighton: This is dope that you set this up as an example/proof of concept but it probably doesn't make sense for this to land on master yet since we are taking the component first approach. We don't need a page level grid yet. I would actually prefer to not even begin to put the components into a layout. I think building them without the layout breaks our work into smaller pieces, gives us less to consider at one time, and encourages us to build the components in a truly reusable fashion

imports/ui/components/AlbumsPage/AlbumsPage.scss
```diff
@@ -0,0 +1,21 @@
+@import "{wolves:bourbon}/bourbon";
```
rosscreighton: We shouldn't have to do these imports in partial level stylesheets. We should really only be importing the libraries once

imports/ui/components/Page/Page.scss
```diff
@@ -0,0 +1,16 @@
+@import "{wolves:bourbon}/bourbon";
+$visual-grid: true;
+$visual-grid-color: yellow;
+$visual-grid-index: front;
+$visual-grid-opacity: 0.5;
+@import "{wolves:neat}/neat";
+
+.page {
+  .container {
+    @include outer-container();
```
rosscreighton: Should we just include outer-container on .Page?

imports/ui/components/Page/Page.scss
```diff
@@ -0,0 +1,16 @@
+@import "{wolves:bourbon}/bourbon";
```
rosscreighton: I feel like these imports and vars belong in some sort of neat-config stylesheet. Do you know how neat recommends setting this up?

package.json
```diff
@@ -29,6 +29,13 @@
...
     "eslint-plugin-meteor": "^3.4.0",
-    "eslint-plugin-react": "^4.2.3"
+    "eslint-plugin-react": "^4.2.3",
+    "autoprefixer": "^6.3.5",
+    "postcss-easy-import": "^1.0.1",
+    "postcss-nested": "^1.0.0",
+    "postcss-simple-vars": "^1.2.0",
+    "postcss-conditionals": "^2.0.2",
+    "postcss-mixins": "^4.0.2",
+    "rucksack-css": "^0.8.5"
```
rosscreighton: Rucksack?

package.json
```diff
@@ -29,6 +29,13 @@
     "eslint": "^2.6.0",
     "eslint-config-airbnb": "^6.2.0",
     "eslint-plugin-meteor": "^3.4.0",
-    "eslint-plugin-react": "^4.2.3"
+    "eslint-plugin-react": "^4.2.3",
+    "autoprefixer": "^6.3.5",
+    "postcss-easy-import": "^1.0.1",
```
rosscreighton: Why all the postcss stuff?

package.json
```diff
@@ -29,6 +29,13 @@
     "eslint": "^2.6.0",
     "eslint-config-airbnb": "^6.2.0",
     "eslint-plugin-meteor": "^3.4.0",
-    "eslint-plugin-react": "^4.2.3"
+    "eslint-plugin-react": "^4.2.3",
+    "autoprefixer": "^6.3.5",
```
rosscreighton: Do we need autoprefixer I we're using bourbon?

sameckmeier: huh, i thought i had removed all the postcss dependencies. either way, removing them.

imports/ui/components/Page/Page.scss
sameckmeier: bourbon recommends importing in the stylesheets. i spent a little bit trying to figure out how to set up a global scss file, but i wasnt able to get it to work. im gonna do a little more research into it

imports/ui/components/Page/Page.scss
sameckmeier: i dont think so because the following classes would adopt those styles

imports/ui/components/AlbumsPage/AlbumsPage.jsx
sameckmeier: right, i agree. thanks for reminding me about this, wanted to delete before pushing to master.

imports/ui/components/Page/Page.scss
rosscreighton: oh yeah we should definitely import in a stylesheet, i just think this is not the right place to do it.

rosscreighton: i think you can just put a `main.scss` file in `/client` and import everything from there.

rosscreighton:
closing this PR

PR #14, opened 2016-07-16
Sass grid

rosscreighton:
per our text discussion


imports/ui/stylesheets/index.scss
```diff
@@ -0,0 +1,4 @@
+@import "{wolves:bourbon}/bourbon";
```
sameckmeier: we could even have a file called libraries and stash bourbon and neat in it and any other libraries we end up using

imports/ui/stylesheets/index.scss
```diff
@@ -0,0 +1,4 @@
...
+@import "{wolves:neat}/neat";
+
+@import "./components/index.scss";
```
sameckmeier: i like this, and seems like the best patch over until, as we said, a more developed meteor implementation of css modules comes out. i dont think this work flow is gonna be that much of a headache. i like how modularized this is. plus, this feels very meteor-y 👍

sameckmeier:
looks good to me and seems to be the best workflow for until a more developed css modules comes out

imports/ui/stylesheets/index.scss
rosscreighton: yeah good idea. idk if we'll end up having enough css libraries to justify a separate file, but if we do should do that.

imports/ui/stylesheets/index.scss
rosscreighton: yep, agreed

PR #15, opened 2016-07-16
Simplify ui directory

rosscreighton:
i'm proposing a change for something that has been bugging me. here are my complaints:

1. all of the subdirectories we had under `ui` were _just react components_. therefore, i think all of these components can go under the `components` directory.

2. the `ui` directory is a little complicated and no fun to look at.

3. our project is not big enough to justify break out these react components on such a granular level.

i'd like to clean this stuff up per the changes in this pull request. what do you think?


sameckmeier:
this looks good. i like the change of removing the containers dir and putting those components in their corresponding presentational container dir. we should think whether we'll want to have a dir for each component given that we aren't, at least in the short term, putting stylesheets and other related files in them, especially for those that we'll be using across multiple pages.

rosscreighton:
yeah thats a good point. i think we do need the directory for each component though as we are still keeping test files with each component.

PR #17, opened 2016-07-26
Redux

sameckmeier:
the default store state values for album and spotifyAlbum are null, so if ya wanta check this out in action, you'll have to change em. also, i havent added meteor methods that will wrap dispatches (thats why we're using redunx thunk cause it allows us to send dispatches async) ill throws these in once we connect this branch with the data HOC branch


package.json
```diff
@@ -30,7 +30,11 @@
     "react": "^15.2.1",
     "react-addons-pure-render-mixin": "^15.2.1",
     "react-dom": "^15.2.1",
-    "react-router": "^2.5.2"
+    "react-router": "^2.5.2",
+    "redux": "^3.5.2",
+    "react-redux": "^4.4.5",
+    "redux-thunk": "^2.1.0",
+    "redux-logger": "^2.6.1"
```
rosscreighton: i've found redux-logger to be pretty unhelpful at tsf. i much prefer the redux dev tools extension. happy to keep it in the project if you like it. just throwing it out there - not sure if you checked out the dev tools extension yet.

imports/lib/redux/reducers.js
```diff
@@ -0,0 +1,38 @@
+import { combineReducers } from 'redux'
+import { SET_SPOTIFY_ALBUM, SET_ALBUM } from './store_types';
```
rosscreighton: not sure I understand the `store_types` naming. these are action types.

imports/lib/redux/reducers.js
```diff
@@ -0,0 +1,31 @@
...
+    case SET_ALBUM:
+      return Object.assign({}, state, {
+        album: album
+      })
+    default:
+      return state;
+  }
+};
+
+const spotifyAlbum = (state = null, action) => {
```
rosscreighton: i think initial state should be of the same type as normal state. here we start out with `null`, which will later become and object with key/values. i think the problem this creates is that we'll have to do type checking inside our components as they are rendering. i.e. `if someProp { _render something _ } else { renderSomething different}`

rosscreighton: i saw that you set up initial state in an earlier commit but got rid of it, can you explain?

imports/lib/redux/reducers.js
```diff
@@ -0,0 +1,31 @@
+import { combineReducers } from 'redux'
+import { SET_SPOTIFY_ALBUM, SET_ALBUM } from './store_types';
+
+const album = (state = null, action) => {
```
rosscreighton: i'd prefer to give these functions a Reducer suffix i.e. `albumReducer`, `spotifyAlbumReducer`

rosscreighton: oh maybe you did this because redux sets the name of the key in state based on this function name?

imports/lib/redux/reducers.js
```diff
@@ -0,0 +1,31 @@
...
+  switch (action.type) {
+    case SET_ALBUM:
+      return Object.assign({}, state, {
```
rosscreighton: i'm pretty sure we don't need to use Object.assign here. i think the way combineReducers works is that each reducer controls only its own branch of the state tree. in other words, within the `album` reducer function, the `state` parameter it receives is only `{ album: {} }`, not  `{ album: {}, spotifyAlbum: {} }`, and vice-versa. so for SET_ALBUM and SET_SPOTIFY_ALBUM, we don't need to preserve any of the old state, we just want to replace it with a new album, so we don't need to merge the old state object into a new object.

imports/lib/redux/reducers.js
```diff
@@ -0,0 +1,31 @@
...
+      return state;
+  }
+};
+
+const reducers = combineReducers({
+  album,
+  spotifyAlbum,
+});
+
+export default reducers
```
rosscreighton: this might be overkill but i like to keep things more terse so i would do:

```
export default combineReducers({
    album,
    spotifyAlbum,
});

File: imports/lib/redux/store.js [hunk 7]
```diff
@@ -0,0 +1,11 @@
...
+import createLogger from 'redux-logger';
+import ReduxThunk from 'redux-thunk';
+import reducers from './reducers'
+
+const logger = createLogger();
+const middleware = [ReduxThunk, logger];
+
+const Store = createStore(reducers, {}, applyMiddleware(...middleware));
+
+export default Store;
```
rosscreighton: again, preferring compactness, i would do:

`export default createStore(reducers, {}, applyMiddleware(...middleware));`

feel free to veto this though

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -1,51 +1,21 @@
...
-    const { album, spotifyAlbum } = this.props;
-
-    return(
-      <div>
-        <div>{album.name}</div>
-        <img src={spotifyAlbum && spotifyAlbum.images[0].url} />
-      </div>
-    )
-  }
+function AlbumPage(props) {
```
rosscreighton: because `props` is just an object, thanks to es6 we can unpack props right when the function receives them as args like so:

`function AlbumPage({ album, spotifyAlbum }) {...`

i find this to be super nice and i'd definitely prefer to do it this way if you're down.

imports/ui/pages/AlbumsPage.jsx
```diff
@@ -0,0 +1,22 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+
+export default class AlbumsPage extends Component {
```
rosscreighton: wait, we already have an `AlbumPage` component. is this supposed to be here?

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -1,51 +1,21 @@
...
-    spotifyAlbum: spotifyAlbum.get(),
-  };
-}, AlbumPage);
+    album: state.album,
+    spotifyAlbum: state.spotifyAlbum
+  }
+}
 
-export default ConnectedAlbumPage;
+export default connect(mapStateToProps)(AlbumPage);
```
rosscreighton: i like to write `mapStateToProps` using arrow functions and a more compact syntax, like so:

```
export default connect( state => {
    album: state.album,
    spotifyAlbum: state.spotifyAlbum
})
```

if we want to get crazy i think we can even do it like this using the object assignment shorthand syntax:

```
export default connect( state => {
    state.album,
    state.spotifyAlbum
})
```

package.json
```diff
@@ -30,7 +30,11 @@
     "react": "^15.2.1",
     "react-addons-pure-render-mixin": "^15.2.1",
     "react-dom": "^15.2.1",
-    "react-router": "^2.5.2"
+    "react-router": "^2.5.2",
+    "redux": "^3.5.2",
+    "react-redux": "^4.4.5",
+    "redux-thunk": "^2.1.0",
```
rosscreighton: we don't use redux-thunk anywhere yet, so it would probably be good to leave it out for now. although i haven't looked at the data loader container thing yet so maybe you're using it there, in which case its cool.

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: AlbumPage is missing prop types validation

rosscreighton:
does the documentation recommend putting all the redux stuff in the `lib` directory? this feels weird to me. it feels like we should change `/ui` to `/frontend` and redux stuff should live somewhere like `imports/frontend/store.js`, `imports/frontend/reducers.js`, etc. like i don't even think these files even need their own directory within `/frontend` because they are so fundamental to the frontend itself.

rosscreighton:
do we need tests for our reducers? we should talk about it but i'm almost positive the answer is yes. thankfully they should be really easy to test because they are just pure functions. we just have to make sure they return the proper state when a given action is received.

rosscreighton:
overall this looks really good dude. i left a ton of comments and a few of them are blocking so we should probably talk or go through another round of comments before we merge this one.

rosscreighton:
I'm having second thoughts about the frontend directory. There's no way to put all of our frontend code in one directory because all the meteor code Is isomorphic and we have frontend code that really does belong in lib

PR #19, opened 2016-07-26
Review component

rosscreighton:
this just creates a review component that renders html - no styling, no layout, not hooked up to album page. i thought this would be a good baby step to get on master, so the two of us stay on the same page as far as how we're going to build components. once your redux work gets on maser, i'll refactor the test and hook this up to the album page. once i can visually see the component on the album page, i can do the layout and styling.

p.s. i was originally taking a different direction with this branch, so you'll see FeedItem in some of the commit messages. FeedItem gets changed to Review in later commits


.eslintrc
```diff
@@ -0,0 +1,14 @@
+---
+extends: eslint-config-airbnb
```
sameckmeier: 👍 eslintrc FTW

imports/ui/components/Review/Review.jsx
```diff
@@ -0,0 +1,29 @@
...
+  );
+}
+
+Review.propTypes = {
+  authorName: PropTypes.string,
+  authorAvatarUrl: PropTypes.string,
+  date: PropTypes.object,
+  rating: PropTypes.number,
+  text: PropTypes.string,
+}
```
sameckmeier: just to clarify, once we push redux to master, are we gonna connect() these components? I like the idea of pushing all data needed for a page inside a page component and then connecting each individual component on the page. but, if you thoughts have changed, and based on how this is set up, are you thinking that we want to have families of components with small component depth and connect the top level component and then pass those props down?

sameckmeier: but otherwise this looks nice 👍

rosscreighton:
oh yeah, once redux is available this component will be connected. it will receive the exact same props via connect so it really won't change much besides exporting the connected component.

PR #20, opened 2016-07-26
Data loading wrapper

sameckmeier:
so after ya check this out, im gonna merge this branch and the redux branch into a new branch and add meteor methods that dispatch actions, which will then be merged into master 👍


sameckmeier:
just added some notes and pseudo for further illustration of thoughts 👍

sameckmeier:
adds subscription and autorun args obj generators

sameckmeier:
Made these tweaks based on all that we've touched on. The general idea is that a component can be functional or not and can generate a HOC called MeteorLoader, passing it a component and an obj with keys autorun and/or subscribe. This HOC injects prototypes, Autorunner and Subscriber, that call start and stop within componenDidMount and componentWillUnmount, respectively.

imports/lib/Subscriber.test.js
```diff
@@ -0,0 +1,45 @@
+if (Meteor.isClient) {
```
sameckmeier: this seems to be a necessary evil as it looks like subscribe is not defined on Meteor when on the server

rosscreighton: i don't mind this. this test shouldn't run on the server anyway, Subscriber is only used on the client.

imports/ui/components/Page/Page.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-export default function Page({ children }) {
+export default Page = ({ children }) => {
```
rosscreighton: i would stick with `function Page` here.
`Page = () =>` creates a function assigned to a variable in the _global_ scope. no bueno.
also, es6 modules doesn't let you `export default` a `const` in the same line that it is defined. you can `export` a const, but not `export default`. not sure why. so that means that we can't do `export default const Page = () =>`. we could do something like:

```
const Page = () => {}
export default Page
```

but that just gets a little wordy. i would stick with `export default function Page` because i think it makes the most sense and that is how the rest of the project is set up at this point.

imports/ui/components/Review/Review.jsx
```diff
@@ -1,29 +1,45 @@
 import React, { PropTypes } from 'react';
+import { connect } from 'react-redux';
 
-export default function Review({
+export default Review = ({
```
rosscreighton: same here

imports/ui/components/Review/Review.jsx
```diff
@@ -1,29 +1,45 @@
...
   authorName: PropTypes.string,
   authorAvatarUrl: PropTypes.string,
   date: PropTypes.string,
   rating: PropTypes.string,
   text: PropTypes.string,
+  album: PropTypes.object,
+  spotifyAlbum: PropTypes.object,
 };
+
+export default connect(mapStateToProps)(Review);
```
rosscreighton: there are two default exports in this component

imports/lib/MeteorLoader/MeteorLoader.jsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    componentWillUnmount() {
+      this.composers.forEach(c => c.stop())
+    }
+
+    render() {
+      return (
+        <WrappedComponent
+          {...this.props}
```
rosscreighton: don't need to pass props and state here

.meteor/packages
```diff
@@ -4,18 +4,18 @@
 # 'meteor add' and 'meteor remove' will edit this file for you,
 # but you can also edit it by hand.
 
-meteor-base             # Packages every Meteor app needs to have
-mobile-experience       # Packages for a great mobile UX
-mongo                   # The database Meteor supports right now
-reactive-var            # Reactive variable for tracker
-tracker                 # Meteor's client-side reactive programming library
+meteor-base@1.0.4             # Packages every Meteor app needs to have
```
rosscreighton: are we sure we want to lock in version numbers? things move pretty quickly in meteor/react land, we'll probably want to upgrade pretty regularly

imports/lib/Autorunner/Autorunner.test.js
```diff
@@ -0,0 +1,33 @@
...
+  const autorunner = new Autorunner(args);
+
+  it('should properly wrap a function in an autorun', function () {
+    autorunner.start();
+    computeAutorun.set(true);
+
+    chai.expect(testVal).to.be.true;
+  })
+
+  it('should stop handles', function () {
```
rosscreighton: justs a suggestion on wording: 'stops handles when Autorunner.stop() is called'

imports/lib/Autorunner/Autorunner.test.js
```diff
@@ -0,0 +1,33 @@
...
+    testVal = true;
+  };
+
+  const args = {
+    test: testFunc
+  };
+
+  const autorunner = new Autorunner(args);
+
+  it('should properly wrap a function in an autorun', function () {
```
rosscreighton: wording suggestion: 'autoruns a function when it's reactive dependency changes'

client/main.js
```diff
@@ -1,6 +1,6 @@
 import '../imports/startup/client';
 import { render } from 'react-dom';
-import { renderRoutes } from '../imports/startup/client/routes';
+import renderRoutes from '../imports/startup/client/routes';
```
rosscreighton: 👍

imports/lib/MeteorLoader/MeteorLoader.test.jsx
```diff
@@ -0,0 +1,27 @@
+/* eslint-env mocha */
```
rosscreighton: i've been using the file naming `.tests.js`. you've been using `.test.js`. i think the plural might make a little more sense because there are multiple test cases per file, but we should just pick one and keep it consistent

imports/lib/Subscriber/Subscriber.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it('should properly subscribe to a subscription', function () {
+      subscriber.start();
+
+      const handle = subscriber.subscriptionHandles.test;
+
+      chai.expect(handle.ready()).to.be.true;
+    })
+
+    it('should stop handles', function () {
```
rosscreighton: my wording suggestions from above also apply here

imports/lib/Subscriber/Subscriber.test.js
```diff
@@ -0,0 +1,45 @@
...
+      subscriber.stop();
+
+      const handle = subscriber.subscriptionHandles.test;
+
+      chai.expect(handle).to.be.null;
+    })
+  })
```
rosscreighton: we should also test if the callback is run as expected

imports/ui/components/Review/Review.jsx
```diff
@@ -1,29 +1,45 @@
...
       <div className="date">{date}</div>
+      <div className="spotifyAlbum">{spotifyAlbum.id || 'loading'}</div>
+      <div className="album">{album.name || 'loading'}</div>
       <div className="rating">{rating}</div>
       <div className="text">{text}</div>
     </div>
   );
 }
 
+function mapStateToProps(state) {
```
rosscreighton: i think we decided a few PR's ago to write the mapStateToProps function inline with connect as an arrow function

imports/ui/components/Page/Page.jsx
sameckmeier: ah good to know about how that scopes it globally. the reason for this change was react-redux wasnt working properly with defining components a la functions. im not entirely sure why, maybe it needs to lexically scope this? there could be something else at hand causing this using-function bug, but that change to the arrow def made everything work.

sameckmeier: so we could change Page back to function, but maybe we want to set a const at the top and then export default for consistency sake? let's chat about this more tomorrow.

imports/ui/components/Review/Review.jsx
sameckmeier: haha good catch

.meteor/packages
sameckmeier: huh weird, i agree that we dont. i believe this happened when updating with patch -- the reason for that patch was that 1.4 broke user-accounts so im guessing MDG set those packages as a means to fix it. will roll this back. we probably shouldnt use 1.4.0.1

imports/ui/components/Review/Review.jsx
sameckmeier: sorry, we did talk about it, but im wondering for sake of documentation to implement this way?  it's possible that we'll want to mapDispatchToProps, another arg to connect, and it seems weird to pass two inline arrow functions. chat a little more about this tomorrow?

rosscreighton: cool yeah lets chat on this tomorrow

PR #22, opened 2016-07-30
Review component

rosscreighton:
styles the Review component.

punting a few decisions for now, we'll need to talk about these soon:

- global variables for default margins, paddings, font-sizes
- fonts. setting up project wide fonts is a big enough story to warrant its own PR.
- not sure how this component should look in responsive states, so i kept it the same as the desktop view. everything flows down fine, but we may want to revisit this later.


rosscreighton:
@sameckmeir when you were researching CSS stuff, did you find a better way to add normalize to the project? The way I did it here seems just ok.

rosscreighton:
There is a meteor package for it but it's out of date. We could use an npm package but I feel like since we're using meteor's build system, things that have to do with the build system (sass processor, sass packages in this case) should be installed via a meteor package. We could package normalize ourselves?? Wouldn't be too hard. Idk it just feels like a lot to check that big CSS library into version control.

imports/ui/stylesheets/index.scss
```diff
@@ -1,3 +1,4 @@
+@import "./vendor/normalize.scss";
```
sameckmeier: although it's not totally ideal, https://guide.meteor.com/using-npm-packages.html#npm-styles, it looks like we should be able to use npm for this and not run into any problems. it doesnt seem great to have normalize bundled with the app/version controlled. there's also this package, https://atmospherejs.com/johnantoni/meteor-normalize. but im down to keep it as is if you think that it is our best option.

imports/ui/components/Review/Review.jsx
```diff
@@ -3,27 +3,44 @@ import React, { PropTypes } from 'react';
 export default function Review({
```
sameckmeier: this all looks good to me! i like the way we're organizing our styles, and your classNames are 👍

imports/ui/stylesheets/index.scss
rosscreighton: @sameckmeier yeah the atmosphere package isn't being maintained :/. but yeah you're right using the npm package isn't a big deal. i'll make the change

imports/ui/components/Review/Review.jsx
rosscreighton: thanks!

PR #23, opened 2016-07-30
Frosted pane

rosscreighton:
"FrostedPane" component i sent you a screen shot of earlier (definitely open to other naming ideas).

- updates meteor to 1.4
- creates helper for setting up jsdom. i looked for a way to do this as a velocity or mocha configuration but i couldn't find anything that wasn't hacky :/. what do you think? is this horrible?


imports/ui/stylesheets/index.scss
```diff
@@ -5,3 +5,7 @@
 @import "./variables.scss";
 
 @import "./components/index.scss";
+
+html {
```
rosscreighton: @sameckmeier should we put this somewhere else or is it ok here for now?

imports/ui/components/FrostedPane/FrostedPane.tests.js
```diff
@@ -0,0 +1,28 @@
...
+
+  const wrapper = mount(
+    <FrostedPane>
+      <TestComponent />
+    </FrostedPane>
+  );
+
+  it('renders its children inside of its wrapper div', function () {
+    chai.expect(wrapper.find('.FrostedPane').contains(<TestComponent />)).to.equal(true);
+  });
```
rosscreighton: @sameckmeier is there anything else about this component that i should test?

imports/ui/stylesheets/variables.scss
```diff
@@ -5,3 +5,15 @@ $smoke: #f1f1f1;
 $mist: #f4f4f4;
 $white: #ffffff;
 
+// amounts
```
sameckmeier: 👍

imports/ui/stylesheets/index.scss
sameckmeier: we could stash this in something like global.scss. it seems that we should keep index.scss as a place where we only load files

imports/ui/components/FrostedPane/FrostedPane.tests.js
```diff
@@ -5,7 +5,9 @@ import React, { Component } from 'react';
 import { chai } from 'meteor/practicalmeteor:chai';
 import { mount } from 'enzyme';
 import FrostedPane from './FrostedPane';
-import jsdom from 'jsdom';
+import { initJSDom } from '../../../lib/testHelpers';
+
+initJSDom();
```
sameckmeier: could we build a test config file that gets executed at meteor startup? we could use the Meteor.isTest flag to check.

imports/ui/stylesheets/index.scss
sameckmeier: we could stash this in something like global.scss. it seems that we should keep index.scss as a place where we only load files

sameckmeier:
overall, looks good, and i like the what FrostedPane is doing. do we want to supply it a className props so that we can make it more reusable, or are you thinking that its better to have multiple, specific semantic design elements.

imports/ui/components/FrostedPane/FrostedPane.tests.js
sameckmeier: we could also make sure that FrostedPanes child adopts FrostedPane's styles

sameckmeier:
just a thought, we could call this StylesWrapper, and its stylesheet would be a ton of different classes, and we'd then pass it a props called className or something.

imports/ui/stylesheets/index.scss
rosscreighton: @sameckmeier yeah i ended up doing this in a later pull request. good call.

rosscreighton:
@sameckmeier i'm going to put an item in trello for us to talk about this next time we do a call

imports/ui/components/FrostedPane/FrostedPane.tests.js
rosscreighton: FrostedPanes child actually shouldn't adopt FrostedPane's styles. the styles in FrostedPane.scss only apply to FrostedPane's wrapper div

PR #24, opened 2016-07-31
creates SectionHeader component

rosscreighton:
pretty straightforward here. lemme know what you think 😎


sameckmeier:
this looks good 👍 just to make sure that im on the same page as you, the spacing of this component relative to others will be handled by a separate component, ie this component doesnt concern itself with how it should be spaced from the review component below it?

rosscreighton:
@sameckmeier that's correct.

rosscreighton:
@sameckmeier for this specific component though, i think the spacing will be handled by it's parent's stylesheet. for example, something like the following in AlbumPage.scss:

```
.AlbumPage {
  .SectionHeader {
    margin-bottom: $gutter-md;
  }
}
```

PR #25, opened 2016-08-03
ui components

rosscreighton:
keeping all my recent react work in this PR until we have a chance to talk. this also give you a chance to look at how the components work together before we merge them to maser.


imports/ui/stylesheets/variables.scss
```diff
@@ -1,6 +1,6 @@
 // colors
 $midnight: #000000;
```
sameckmeier: im wondering if we want to have a more consistent naming scheme? what if we used crayola color names so that we could reference: https://en.wikipedia.org/wiki/List_of_Crayola_crayon_colors#Standard_colors

that idea came from http://coryetzkorn.com/blog/color-variables-in-sass/ 

im also intrigued by how this guy prefixes all color var names with 'color-'. i can see the benefit but it also seems a little redundant. 

lastly, what are your thoughts on setting functional variable names (eg $border-color) within stylesheets? the thought is that we'd have a colors.scss and then in Reviews.scss we set all of the colors that we use for that component at the top of the stylesheet. again, these are just thoughts, and i havent really decided if they'd be useful or just redundant and tedious.

imports/ui/stylesheets/components/PageHeader.scss
```diff
@@ -1,10 +1,14 @@
 .PageHeader {
-  height: 270px;
-  @include box(column, center, center);
+  $header-height: 270px;
+  $header-padding: 35px;
```
sameckmeier: do we want to call this header-margin?

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -14,16 +14,18 @@ class AlbumPage extends Component {
...
         <PageHeader>
           this is page header content
         </PageHeader>
-        <SectionHeader>
-          Reviews
-        </SectionHeader>
-        <FrostedPane>
-          <Review />
-        </FrostedPane>
+        <div className="body">
```
sameckmeier: nice 👍

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -14,16 +14,18 @@ class AlbumPage extends Component {
     console.log(this.props);
 
     return (
-      <div>
+      <div className="AlbumPage">
         <PageHeader>
```
sameckmeier: would it be redundant to wrap Page header in a div called .header?

imports/ui/stylesheets/index.scss
```diff
@@ -4,8 +4,16 @@
 
 @import "./variables.scss";
 
+@mixin outer-container-narrow() {
```
sameckmeier: what do you think about naming this outer-container-three-fourths?

imports/ui/stylesheets/mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin outer-container-narrow() {
```
sameckmeier: what are your thoughts on naming this outer-container?

imports/ui/stylesheets/mixins.scss
```diff
@@ -0,0 +1,7 @@
...
+  @include span-columns(12);
+  @include media($bp-tablet) {
```
sameckmeier: and then naming the smaller defaults something like three-fourths or something of the like?

imports/ui/components/TruncatedText/TruncatedText.jsx
```diff
@@ -0,0 +1,41 @@
+import React, { Component, PropTypes } from 'react';
+
+export default class TruncatedText extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      truncated: true,
+    };
+    this.toggleTruncate = this.toggleTruncate.bind(this);
```
sameckmeier: 👍 really like this as opposed to binding in render

imports/ui/components/TruncatedText/TruncatedText.jsx
```diff
@@ -0,0 +1,41 @@
...
+    const { truncated } = this.state;
+    const truncatedText = text.split(' ').slice(0, wordCount).join(' ');
+
+    return (
+      <span className="TruncatedText">
+        <span className="text">
+          {truncated ? truncatedText : text}
+        </span>
+        <a href="#" onClick={this.toggleTruncate}>
+          {truncated ? ' MORE...' : ' (show less)'}
```
sameckmeier: would it be redundant to define consts for each of these?

imports/ui/components/TruncatedText/TruncatedText.tests.js
```diff
@@ -0,0 +1,59 @@
...
+    wrapper = mount(
+      <TruncatedText
+        text={testText}
+        wordCount={wordCount}
+      />
+    );
+  });
+
+  it('renders correctly truncated text by default', function () {
+    const words = wrapper.find('.text').text().split(' ').length;
```
sameckmeier: we could define a function for this to dry this up

imports/ui/components/TruncatedText/TruncatedText.tests.js
```diff
@@ -0,0 +1,59 @@
...
+    chai.expect(words).to.equal(wordCount);
+  });
+
+  it('sets truncated state to false when MORE is clicked', function () {
+    wrapper.find('a').simulate('click');
```
sameckmeier: and same for this one

imports/ui/components/TruncatedText/TruncatedText.jsx
```diff
@@ -20,11 +20,11 @@ export default class TruncatedText extends Component {
   render() {
     const { text, wordCount } = this.props;
     const { truncated } = this.state;
-    const truncatedText = text.split(' ').slice(0, wordCount).join(' ');
+    const truncatedText = text.split(' ').slice(0, wordCount).join(' ').trim();
```
sameckmeier: 👍

sameckmeier:
all in all looks good! my only blocking questions/comments are about color and spacing naming conventions.

imports/ui/components/TruncatedText/TruncatedText.jsx
rosscreighton: yeah, `bind` returns a new function with a new context, so if you use bind inside the render method, it creates a new function every time the component re-renders. can cause performance issues

imports/ui/stylesheets/variables.scss
rosscreighton: yeah i like that idea. i was already using the [name that color](http://chir.ag/projects/name-that-color/) app suggested in that article, but i like the idea of having the wikipedia list of colors to reference. 

and yeah i think i like prefixing with color also. that thought has crossed my mind before. 

i'll make these updates.

re: the last paragraph in your comment. i can see how that could be useful, but i'm genuinely not sure if we need to do something like that. my thinking is that _hopefully_ our components are small enough that we don't need to add extra context to variable names for the stylesheet to make sense. i'm happy to try it out though.

imports/ui/stylesheets/components/PageHeader.scss
rosscreighton: yeah good call `$header-gutter-top` or just `$header-gutter` makes more sense.

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: yeah i think so because `<PageHeader />` renders out the wrapper `<div className="PageHeader"></div>`. if for some reason we needed to apply a style to the PageHeader component, we could reference it using the .PageHeader class like:

```
.AlbumPage {
  .PageHeader {
    // some album page specific style for the page header
  }
}
```

imports/ui/components/TruncatedText/TruncatedText.jsx
rosscreighton: good call

imports/ui/components/TruncatedText/TruncatedText.tests.js
rosscreighton: yeah good catch i should definitely dry this up.

rosscreighton:
good stuff man, i'm going to make a few updates to this branch and send it back to you.

PR #30, opened 2016-08-06
switches to Lost grid system, removes bourbon and neat

rosscreighton:


imports/ui/stylesheets/mixins.scss
```diff
@@ -1,3 +1,12 @@
 @mixin outer-container-narrow() {
```
sameckmeier: really like how succinct the logic and naming is 👍

rosscreighton: yeah i agree i think this mixin could definitely use some work. i'm going to add a ticket to trello for us to talk about this. once we get some definition around what the mixin should do, we can implement.

PR #31, opened 2016-08-06
Manual css modules

rosscreighton:
manually scopes classes until we can get css-modules working with our build system


sameckmeier:
i like this convention all things considered. this feels like a solid stop gap until we get css modules working.

sameckmeier:
also, and sorry i just need a refresher, what was the pain point again that precipitated this change ie prepending ComponentName to class names?

rosscreighton:
for sure, mind if i explain it next time we talk? i don't want to type it out haha.

PR #32, opened 2016-08-07
creates HeaderContent component

rosscreighton:


sameckmeier:
looks good! id love to take a minute and chat with you about responsive layouts and how youre setting em up. i just have a few ?s on how it works.

sameckmeier:
:shipit:

sameckmeier:
we should probably talk briefly about how to either make this abstract so that it works on other pages or change the name so that it is specific to the album page

rosscreighton:
cool i'm going to add both of these to trello

rosscreighton:
re: abstracting this component,

so i think i've reached the appropriate level of abstraction. i built this in such a way that it can be shared between ArtistPage and AlbumPage. these are the only two pages that have a header besides the FAQ page, and the FAQ page header content is extrememely simple and doesn't need any weird responsiveness, so it wouldn't need a HeaderContent component. also, the FAQ won't be built for a long time, so i didn't think it made sense to take it into account at the moment.

i do, however, think this component could have a more intention revealing name. any ideas?

rosscreighton:
`ReviewableHeaderContent`? `HeaderContentReviewable`?

PR #33, opened 2016-08-07
removes unnecessary sass nesting

rosscreighton:
we don't need usually need nesting now that we are manually prefixing variable names. sass nesting produces non performant css, so we should avoid it when possible. as a bonus, this is closer to how our style sheets will look when we get css modules working


sameckmeier:
good to know about how nesting isnt performant. all in all, this looks good. i like how modular it is and that it's evident where a components global styles go 👍

sameckmeier:
:shipit:

PR #34, opened 2016-08-07
creates Navbar component

rosscreighton:


imports/ui/stylesheets/mixins.scss
```diff
@@ -1,3 +1,16 @@
+@mixin outer-container-wide() {
```
sameckmeier: per a comment on a previous PR, what are your thoughts on using a more specific size naming? i think this works but am just wondering if we can be more specific.

PR #35, opened 2016-08-08
sets up fonts

rosscreighton:
i added more fonts to the project than we actually need. i just wanted to move them all over from the old repo. once we get a little further along with the design, we can remove the ones we don't need. i added a ticket to trello for this.


sameckmeier:
:shipit:

PR #40, opened 2016-08-16
MeteorLoader integration

rosscreighton:


imports/redux/actions.js
```diff
@@ -1,9 +1,11 @@
 import { SET_SPOTIFY_ALBUM, SET_ALBUM } from './action-types';
 
-export function setSpotifyAlbum(spotifyAlbum) {
-  return { type: SET_SPOTIFY_ALBUM, spotifyAlbum };
-}
+export const setSpotifyAlbum = (spotifyAlbum) => ({
```
sameckmeier: i like that you did this. i just need to get it in my head that in situations like this, we should be using arrow functions. i want to make similar changes to spotify 👍

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -47,25 +48,34 @@ AlbumPage.propTypes = {
...
+}
 
-//       HTTP.get(spotifyAlbumQueryUrl, (err, res) => {
-//         const spotifyAlbum = formatRawAlbum(res.data);
-//         dispatch(setSpotifyAlbum(spotifyAlbum));
-//       });
-//     },
-//   },
-// };
+export default connect()(AlbumPageLoader);
```
sameckmeier: ahh ok i see what the problem was with this. we only run into the infinite loop when we connect AlbumPage to our store. just to throw this out there again, an alternative to this is that we make it so that that AlbumPage doesnt  pass props down to any components it renders. this would also cause HeaderContent to connect to album and artist stores thereby making it so that it never receives props from the component that renders it. 

as we touched on, this would mean that for components like Review, it'd need to be wrapped in a Reviews or Feed component. To make that component abstract, we could have AlbumPage dispatch an action that'd set its props and tell it how and what to render. 

All in all, if you feel strongly about not going in that direction, as we talked about on Sunday, then im cool with going this route -- i think these changes look good and like how it's a bit more explicit. I just think that the primary advantage of the setup i listed above is that responsibilities are truly separated: Page components hydrate the store, redux passes props into components, components connect to redux and are purely functional, and all is modular. this would also makes it easier to rinse and repeat when building components ie we dont need to wonder if we need to connect it etc.

imports/redux/reducers.js
```diff
@@ -31,6 +31,8 @@ const album = (state = {}, action) => {
 
   if (action.type === SET_ALBUM) {
     newState = action.album;
+  if (action.type === SET_SPOTIFY_ALBUM) {
+    newState = Object.assign({}, state, { imageUrl: action.spotifyAlbum.images[1].url });
```
sameckmeier: this is a good catch and completely agree! 👍 this does a great job at limiting the reach of spotify.

imports/ui/components/AlbumPage/AlbumPage.jsx
sameckmeier: with that setup though, there are some downsides. we touched on the one the arises from wrapping Review in a Reviews/Feed component. the other one is that we'd be leaning on duck typing for a lot of components. For example, HeaderContent would be aware of the inner-workings of the album and artist object, or in other words, would expect an object -- in the case of album -- to be able to receive imageUrl.

imports/redux/reducers.js
```diff
@@ -26,11 +26,13 @@ const artist = (state = {}, action) => {
   return state;
 };
 
-const album = (state = {}, action) => {
+const album = (state = { reviewCount: 20, averageRating: 8.4 }, action) => {
```
sameckmeier: 👍

PR #42, opened 2016-08-17
Test env

sameckmeier:
this gets us closer to automated testing. the only part of this setup that isnt ideal is that whenever a file changes, all of our tests execute. ive yet to find a package that allows you to selectively run tests when a file changes. do you know of any such tool? we should also talk about setting up a CI workflow for down the line, though, i feel that this setup does the equivalent in the meantime.


rosscreighton:
❗ i pulled down this branch and started the tests in watch mode. the tests seem to be working great, but when i try to open the app in a browser there are js errors that prevent it from loading. does this happen to you too?

> this gets us closer to automated testing. the only part of this setup that isnt ideal is that whenever a file changes, all of our tests execute. ive yet to find a package that allows you to selectively run tests when a file changes. do you know of any such tool? 

i don't, i just assumed thats how watch mode would work :/. oh well, i guess we'll wait for that feature.

> we should also talk about setting up a CI workflow for down the line, though, i feel that this setup does the equivalent in the meantime.

absolutely. i think CI will be an important project when we start preparing to actually deploy this thing.

sameckmeier:
good catch! we just needed to change the port for our test env, which is what the recent commit is 👍

rosscreighton:
oh nice. :shipit:

rosscreighton:
wait, mine still isn't working. so you can `npm run tests` and then navigate to localhost:3030/album/bad-blood and the page loads? i still get js errors

rosscreighton:
@sameckmeier

PR #43, opened 2016-08-21
sets up artist data

rosscreighton:
- hooks up artist collection to album collection
- updates databaseSeeder to handle creation of a related artist when we seed an album
- sets up redux stuff for setting the artist in the store after we subscribe to the album publication


imports/api/Artists/server/publications.js
```diff
@@ -0,0 +1,3 @@
+import { Artists } from '../Artists';
+
+Meteor.publish('artist', (id) => Artists.find({ _id: id }));
```
rosscreighton: i set up this publication but i didn't actually need it yet. see comment i left on albums publication.

imports/api/Albums/server/publications.js
```diff
@@ -1,3 +1,9 @@
 import { Albums } from '../Albums';
```
rosscreighton: @sameckmeier take a look at what's happening here and in AlbumPageLoader. let me know if you don't follow why i did this and i'll explain it over the phone. i think we may want to consider a small design change to meteor loader so we don't have to do things like this.

imports/api/artists/Artists.js
```diff
@@ -1,13 +1,13 @@
...
 
-const Artists = new Mongo.Collection('Artists');
+export const Artists = new Mongo.Collection('Artists');
 
 Artists.schema = new SimpleSchema({
   name: { type: String },
   genre: { type: Number },
   spotifyId: { type: String },
-  spotifySlug: { type: String },
+  slug: { type: String },
```
sameckmeier: 👍

imports/lib/database.js
```diff
@@ -8,16 +10,45 @@ const documentsToSeed = [
...
+        newAttributes[attribute] = relatedId;
+      }
+    }
+  }
+
+  return newAttributes;
+}
+
 export function seedDatabase() {
+  resetDatabase();
```
sameckmeier: 👍

imports/redux/reducers.js
```diff
@@ -27,8 +27,14 @@ export const defaultReviewsState = [
   },
 ];
 
-const artist = (state = {}, action) => {
-  return state;
+const artist = (state = { name: 'NAO'}, action) => {
+  let newState = state;
+
+  if (action.type === SET_ARTIST) {
```
sameckmeier: we'll also need to setup a spotify artist action/reducer

imports/api/Albums/server/publications.js
sameckmeier: just to make sure, we covered this over the phone per our conversation about changes to MeteorLoader, ie removing Subscriber, so we can change this back to only returning an album?

.eslintrc
```diff
@@ -12,3 +12,7 @@ globals:
 rules:
   import/no-unresolved: off
   no-underscore-dangle: off
+
+parserOptions:
+  ecmaFeatures:
+    experimentalObjectRestSpread: true
```
sameckmeier: 👍

sameckmeier:
@rosscreighton looks good! the only blocker is that we'll want to remove the patch for the album pub per our conversation on Sunday.

imports/api/Albums/server/publications.js
rosscreighton: for sure, and i'll push the changes for this once i get the tests working 👍

rosscreighton:
@sameckmeier alright its all going to master!!

PR #45, opened 2016-08-22
Compose review

sameckmeier:
i commented out the dispatch code but im hoping that its still enough to get an idea of it. there's also some defaultProps that'll get deleted once we're ready to merge this. also, i just remembered that ill need to add the avatar/profile pic.


imports/ui/stylesheets/colors.scss
```diff
@@ -4,3 +4,5 @@ $color-timberwolf: #d9d9d9; // timberwolf
 $color-baby-powder: #f1f1f1; // baby-powder
 $color-cultured-pearl: #f4f4f4; // cultered-pearl
 $color-white: #ffffff;
+$color-smokey-topaz: #832A0D; // smokey-topaz
```
rosscreighton: i don't know why i left those comments to the right of the the variables. you don't need to to this.

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -16,26 +17,13 @@ function AlbumPage({
 }) {
   return (
     <div className="AlbumPage">
-      <PageHeader>
```
rosscreighton: why was this block of code deleted?

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -16,26 +17,13 @@ function AlbumPage({
...
-            { label: 'reviews', value: album.reviewCount },
-            { label: 'average rating', value: album.averageRating },
-          ]}
-        />
-      </PageHeader>
       <div className="AlbumPage-body">
         <SectionHeader>
           Reviews
         </SectionHeader>
-        {reviews.map((review, i) => (
```
rosscreighton: same thing here. why are we deleting HeaderContent and reviews?

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -0,0 +1,121 @@
...
+        return (
+          <button className={ratingClassName} onClick={this.handleRatingClick} key={i}>{rating}</button>
+        );
+      }));
+    };
+
+    return (
+      <div className="ComposeReview">
+        <div className="ComposeReview-side">
+          <div className="ComposeReview-avatar">
```
rosscreighton: we could _maybe_ make an argument for creating an avatar component. what do you think

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -0,0 +1,121 @@
...
+    });
+  }
+
+  render() {
+    const { ratings, defaultText, userAvatarUrl, userName } = this.props;
+    const { selectedRating, text, validText, validRating } = this.state;
+
+    const textClassName = validText ? '' : 'ComposeReview-invalid';
+
+    const renderRatings = (ratings) => {
```
rosscreighton: i definitely wouldn't define this function inside of `render()`, because the function gets recreated in memory every time the component re-renders, which can lead to performance issues.

rosscreighton: we should talk about this and get on the same page, but at a higher level, i think its usually a bad idea to have functions besides the render method of a component that produce jsx. i think its really nice to have the render method read like an html document and have it flow from start to finish without breaking out into other functions, causing the reader to jump to another place in the file. i just think it makes things easier to read. i've found that its almost always possible to avoid moving piece of the jsx into other functions.

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -0,0 +1,121 @@
...
+  albumName: PropTypes.string.isRequired,
+  albumId: PropTypes.string.isRequired,
+  userId: PropTypes.string.isRequired,
+  ratings: PropTypes.array,
+  defaultText: PropTypes.string,
+  userAvatarUrl: PropTypes.string,
+};
+
+ComposeReview.defaultProps = {
+  ratings: ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'],
```
rosscreighton: not sure which is the right way to go, but the Artists collection schema defines the rating attribute as type `Number`. we should decide which one makes the most sense and keep them consistent. i think using numbers instead of strings probably makes more sense as we will want to do math with the ratings. if we can avoid casting strings to the numbers that would be nice.

rosscreighton: should we put this array of rating options in a lib file or something?

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  .ComposeReview-main {
+    margin-right: $gutter-sm;
+
+    @media screen and (min-width: $breakpoint-tablet) {
+      width: 100%;
+    }
+
+    @media screen and (max-width: $breakpoint-mobile) {
```
rosscreighton: so as far as i know, its considered best practice to use min-width breakpoints. set default styles for the mobile viewport, and override them to make changes at different breakpoints as the viewport gets larger. it looks like you went with `max-width` break points for this component. what was your reasoning there? 

whether or not it is good practice, i think its important for us to approach responsive layouts in the same way so the codebase is consistent.

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  .ComposeReview-ratings button {
+    width: $box-size;
+    height: $box-size;
+    background-color: $color-cultured-pearl;
+    color: $color-timberwolf;
+    text-align: center;
+    font-size: rem(18);
```
rosscreighton: we have a variable for this font size. `$font-size-md`

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+    font-weight: bold;
+    margin-right: 10px;
+    border: none;
+
+    &.ComposeReview-selectedRating {
+      background-color: $color-white;
+      border: 1px solid $color-timberwolf;
```
rosscreighton: we should also set the font color to black when a rating is selected so it stands out a little more. switching the background from light grey to white is pretty sudle - i don't think it gives good visual feedback to the user. the mock also shows it changing to black.

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+
+  .ComposeReview-ratings button {
+    width: $box-size;
+    height: $box-size;
+    background-color: $color-cultured-pearl;
+    color: $color-timberwolf;
+    text-align: center;
+    font-size: rem(18);
+    font-weight: bold;
+    margin-right: 10px;
```
rosscreighton: i think we should also give the ratings div a margin-top: 10px, so the spacing in the component is a little more consistent and the rating buttons don't crowd up against the textarea.

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+  .ComposeReview-avatar > img {
+    height: $box-size;
+    width: $box-size;
+    @media screen and (max-width: $breakpoint-mobile) {
+      height: calc(#{$box-size} * 2);
+      width: calc(#{$box-size} * 2);
+    }
+  }
+
+  .ComposeReview-text > textarea {
```
rosscreighton: i think we should give this textarea a few pixels of inner padding so the text doesn't crowd up against the border of the box.

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -0,0 +1,128 @@
...
+    resize: none;
+    color: $color-gray;
+    border: 1px solid $color-timberwolf;
+    height: 100px;
+    width: calc(100% - 6px);
```
rosscreighton: its not clear where `6px` comes from. maybe we could give this some context by putting it in an variable?

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+    if (selectedRating === null) { validRating = false };
+    if (text === null || text.match(/\w/) === null) { validText = false };
+
+    if (validRating && validText) {
+      // dispatch(composeReview(args))
+      // if/else for error handling based on dispatched action relating to review creation error
+      this.setState(this.defaultState);
+    } else {
+      this.setState({
```
rosscreighton: i think the validation and error state belongs in redux. i think always a good idea to limit the amount of state a component has.

rosscreighton:
looks like we do have a default border radius variable in layout-amounts.scss (i don't like my naming choice btw)

i think anything in this component that has a border should get a `border-radius: $border-radius`, including the submit button. i know the mock shows the button having a more rounded corner but i think it looks inconsistent and less modern.

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -0,0 +1,121 @@
...
+            <img src={userAvatarUrl} alt={`${userName}'s avatar`} />
+          </div>
+        </div>
+        <div className="ComposeReview-main">
+          <div className="ComposeReview-text">
+            <textarea className={textClassName} onChange={this.handleReviewChange} value={text} placeholder={defaultText} />
+          </div>
+          <div className="ComposeReview-ratings">
+            <div>
+              {renderRatings(ratings.slice(0, ratings.length/2))}
```
rosscreighton: just throwing this out there, you could've accomplished the rating rows here with css, which is what i would've done, but your way is good too and i think its actually probably a little easier to read.

rosscreighton: but in general i think its better to use css when you can. the browser is much faster at parsing/painting styles than it is a executing javascript and mutating the DOM.

imports/ui/components/ComposeReview/ComposeReview.tests.js
```diff
@@ -0,0 +1,66 @@
+/* eslint-env mocha */
```
rosscreighton: some tests i think we should add:

- clicking a rating button adds the active class to that button.
- clicking the submit button dispatches an action to create the review.

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: nvm, looks like you just wanted to get it out of the way while you prototyped

rosscreighton:
so i want to have a high level discussion with you about how we're going to use redux to manage component state. we've had heated discussions on this at tsf, but i think we've landed on an approach that makes a ton of sense. i want to explain it over the phone, but just to give you an idea, ComposeReview's state would be moved entirely into redux, it would have its own reducer, and the component would be entirely functional.

rosscreighton:
good shit dude! 

the only blocking comments are:
- using the existing variables we have for font sizes and border radiuses
- making a couple design/spaces tweaks (i feel strongly about them, but if you feel differently thats cool 👍 )
- move that one function out of the render method

would be nice to have:
- slightly better test coverage.

and i definitely want to talk about the following before we ship this just so we're on the same page:
- min-width vs. max-width breakpoints
- moving this component's state to redux

imports/ui/components/ComposeReview/ComposeReview.jsx
sameckmeier: ya i like that idea. a component that takes an avatar image (could be an artist, album, user, etc), and we give it a width and height plus a name. ill whip one up 👍

imports/ui/components/AlbumPage/AlbumPage.jsx
sameckmeier: ill comment all of this back in before i merge

imports/ui/components/ComposeReview/ComposeReview.jsx
sameckmeier: mmmmm really good point about render! ill change that. 

ya, let's chat about your render() point. i feel that if a function reads semantically and doesnt confuse what's being rendered, then it shouldnt require someone to think about the implementation. on the flip side, is that line of thinking off in this case and actually makes it tougher to reason about what's happening?

imports/ui/components/ComposeReview/ComposeReview.jsx
sameckmeier: ah, yep, agree with casting ratings as integers.

it's probably good practice to set that array to a const, but if we only use this once then it just might extend the scope of this component. im trying to imagine another instance when we'll need it?

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: this was something i wanted to chat with you about :) and you also answered the question i had, design is mobile first which is one uses min and then override as the viewport gets bigger. ill change this. this was just a means to get it up and running because i didnt entirely understand the reasoning behind min-width, max-width just made more sense. either, im down for min-width!

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: ah right, cool.

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: i thought that the gray looked good, but i agree with your point that black sticks out a little better. i might go with an grayer black just to keep with the color scheme of using subtler colors. thoughts? may do the same with the red we use for invalid entries.

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: 👍

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: 👍

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: mmm good point. the textarea pushes a little past its container by 6px. so im not entirely sure if we need it, as this could be do to something else -- i was having a hard time figuring out why that was the case.

imports/ui/components/ComposeReview/ComposeReview.jsx
sameckmeier: agreed! and good catch. ill go through this and push this components state up to redux. as a rule of thumb, we should indeed limit a component's state and delegate as much as possible to redux 👍

imports/ui/components/ComposeReview/ComposeReview.jsx
sameckmeier: sorry my newb is showing ;) could we chat about how you wouldve done that with css?

imports/ui/components/ComposeReview/ComposeReview.tests.js
sameckmeier: the active class is a good idea. was thinking that once we get the review action/reducer all merged, id come back and write a test for it.

sameckmeier:
fantastic comments/advice!! youre eye is strong 💪 ive added a few discussion points to trello. ill go through and make changes. also, i like the format of your final comment. id like to start using that for future prs. ill message you when im finished with your changes :)

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: i would give it t-shirt sizes: `sm, md, large`. if the component can accept any width/height props, it might encourage us to create inconsistently sized avatars across the site

imports/ui/stylesheets/components/ComposeReview.scss
rosscreighton: i just think the text should be darker when the rating is selected. i'm cool with whatever color . we **should** be cautious about creating too many color options though.

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: fer sher i'll explain on the phone

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -54,22 +54,24 @@ export class ComposeReview extends Component {
     });
   }
 
+  renderRatings(ratings) {
```
rosscreighton: 👍

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -5,13 +5,13 @@
   background-color: $color-white;
   display: flex;
   justify-content: space-around;
+  flex-direction: column;
 
-  @media screen and (max-width: $breakpoint-mobile) {
-    flex-direction: column;
+  @media screen and (min-width: $breakpoint-tablet) {
```
rosscreighton: perfecto 👍

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -57,71 +56,79 @@
...
+      width: $box-size;
     }
   }
 
   .ComposeReview-text > textarea {
+    $rectify-width: 23px;
     resize: none;
     color: $color-gray;
-    border: 1px solid $color-timberwolf;
+    border: $border-radius solid $color-timberwolf;
```
rosscreighton: the first argument to `border` is `border-width`. the above sets border-width to 2px. there is no shorthand property for border-radius, so you have to add it on its own line:

`border-radius: $border-radius`

imports/ui/stylesheets/components/ComposeReview.scss
```diff
@@ -57,71 +56,79 @@
...
+    font-size: $font-size-md;
     font-weight: bold;
+    margin-top: 5px;
     margin-right: 10px;
     border: none;
 
     &.ComposeReview-selectedRating {
       background-color: $color-white;
-      border: 1px solid $color-timberwolf;
+      border: $border-radius solid $color-timberwolf;
```
rosscreighton: same here, we want this to be 1px and set border radius on a different line

imports/ui/stylesheets/components/ComposeReview.scss
sameckmeier: haha brainfarts - changing now

rosscreighton:
@sameckmeier closing this as all work was transferred over to compose-integration

PR #46, opened 2016-08-25
converts redux setup to ‘ducks’ organization

rosscreighton:


rosscreighton:
i feel like calling the directory `ducks` could be misleading bc. but also the directory is inside the redux directory, so it makes a buttload of sense if you're familiar with the redux ducks proposal.

imports/redux/ducks/album.js
```diff
@@ -0,0 +1,45 @@
...
+
+export const defaultAlbumState = {
+  reviewCount: 20,
+  averageRating: 8.4,
+};
+
+export default function reducer(state = defaultAlbumState, action) {
+  let newState = state;
+
+  if (action.type === SET_ALBUM) {
```
sameckmeier: i think we  should turn this into if/elsif so that it doesnt execute anymore than this needs to

imports/redux/ducks/index.js
```diff
@@ -0,0 +1,11 @@
+import { combineReducers } from 'redux';
```
sameckmeier: i like that this explicitly called ducks 👍

sameckmeier:
blocking comments:

- using if/elsif instead of sequential if statements

package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "start": "meteor run",
-    "tests": "TEST_WATCH=1 meteor test --port 3030 --driver-package dispatch:mocha",
+    "tests": "SERVER_TEST_REPORTER=min TEST_WATCH=1 meteor test --port 3030 --driver-package dispatch:mocha",
```
sameckmeier: 👍

rosscreighton:
@sameckmeier addressed comments, merging

PR #47, opened 2016-08-29
ComposeReview integration

rosscreighton:


imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -39,7 +39,7 @@ export function ComposeReview({
       </div>
     </div>
   );
-};
```
sameckmeier: what are your thoughts on when and when not to use semi colons?

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -20,18 +20,38 @@ export function ComposeReview({
         <div className="ComposeReview-content">
           <textarea placeholder="Compose review..." />
         </div>
-        <div className="ComposeReview-ratings">
-          <div>
-            {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10].map((rating, i) => (
+        {(() => {
```
sameckmeier: because this function is pretty long, what are your thoughts on keeping it in that original function? i agree with you on that we want only render to return jsx as much as possible, but it feels like there might be instances where itd be a little more readable if we hid some ui implementation behind a semantic function.

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -20,18 +20,38 @@ export function ComposeReview({
...
                 onClick={() => dispatch(selectRating(rating))}
                 key={i}
               >
                 {rating}
               </button>
-            ))}
-          </div>
+            );
+
+            if (rating <= 5) {
```
sameckmeier: length/2 might be preferable here

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -20,18 +20,38 @@ export function ComposeReview({
         <div className="ComposeReview-content">
           <textarea placeholder="Compose review..." />
         </div>
-        <div className="ComposeReview-ratings">
-          <div>
-            {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10].map((rating, i) => (
+        {(() => {
+          const topRow = [];
+          const bottomRow = [];
+          const ratings = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
```
sameckmeier: let's chat about having a default ui values files for contsts

imports/ui/stylesheets/global.scss
```diff
@@ -11,3 +11,13 @@ img {
   width: 100%;
   height: auto;
 }
+
```
sameckmeier: 👍 nice

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -25,24 +26,24 @@ export function ComposeReview({
...
           </Debounce>
         </div>
         {(() => {
           const topRow = [];
           const bottomRow = [];
-          const ratings = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+          const ratingOptions = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
 
-          ratings.forEach((rating, i) => {
+          ratingOptions.forEach((r, i) => {
```
sameckmeier: 👍

imports/redux/ducks/composeReview.js
```diff
@@ -17,40 +18,60 @@ export const selectRating = rating => ({
...
-  invalidRating: true,
+  invalidRating: false,
+  invalidContent: false,
 };
 
 export default function reducer(state = defaultState, action) {
   switch (action.type) {
     case SET_CONTENT:
-      return Object.assign({}, state, { content: action.content });
+      return Object.assign({}, state, { content: action.content, invalidContent: !action.content.length });
```
sameckmeier: 👍

imports/redux/ducks/composeReview.js
```diff
@@ -17,40 +18,60 @@ export const selectRating = rating => ({
...
     case SELECT_RATING:
-      return Object.assign({}, state, { rating: action.rating });
+      return Object.assign({}, state, { rating: action.rating, invalidRating: false });
```
sameckmeier: 👍

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -79,17 +83,22 @@ export function ComposeReview({
...
 };
 
 ComposeReview.defaultProps = {
   userAvatarUrl: '/images/avatar_default.jpg',
 };
 
 export default connect(state => ({
+  albumId: state.album._id,
   content: state.ComposeReview.content,
+  invalidContent: state.ComposeReview.invalidContent,
```
sameckmeier: just a question, i had thought that state wasnt scoped and that all states lived as top level keys?

imports/redux/ducks/composeReview.js
```diff
@@ -60,6 +60,7 @@ export const defaultState = {
   rating: null,
```
sameckmeier: it feels as if we should make it so that submitSuccess returns false after one is successfully submitted

imports/redux/ducks/composeReview.js
```diff
@@ -72,6 +73,8 @@ export default function reducer(state = defaultState, action) {
       return Object.assign({}, state, { invalidRating: action.invalidRating });
     case SET_INVALID_CONTENT:
       return Object.assign({}, state, { invalidContent: action.invalidContent });
+    case SUBMIT_REVIEW_SUCCESS:
```
sameckmeier: it feels as if we should make it so that submitSuccess returns false after one is successfully submitted, once that Thank you message is rendered

imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -12,72 +12,75 @@ export function ComposeReview({
...
+                content,
+              }))}
+            >
+              Post
+            </button>
+          </div>
         </div>
-      </div>
+      ]}
+      {submitSuccess && <div className="ComposeReview-success">Thank you!</div>}{/* we should come up with a better success state here */}
```
sameckmeier: im wondering if there's a better way to do this? we should have a way to render ComposeReview with submitSucees: false after one successfully is submitted

sameckmeier:
there isnt any thing blocking, but a few discussion points:

render jsx outside of render
default ui file for consts
way to render default ComposeReview after successful review is submitted so that it's not locked at the Thank you message

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: whenever eslint tells me to 😄

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: i don't think i understand the question but what might help you is just to look at the state object in redux dev tools. its just a plain js object, once you see it, it will probably make more sense.

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: yeah i'm not the most pleased with this either. i'm definitely open to suggestions.

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: i went with a hardcoded value bc i assumed the rating options would always be the same (1-10). but you know what they say about assumptions. i'll change it to length/2 👍

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: not sure what you mean but definitely curious to hear your idea

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: yeah definitely hear what you're saying. i'll explain my reasoning and we should definitely talk about our approach for these types of situations. 

i think there is a trade off happening here with our focus on using only functional components:

functional components encourage the use of only a single function because the props passed into the component are only available within the scope of that component function. in stateful class components, props are made available via the constructor and can be accessed by any arbitrary number of methods using `this.props.someProp`. but with functional components, you can't just reference the context to access props, you have to explicitly pass props around as arguments to functions.

in this case, i would have to pass dispatch, albumId, content, invalidContent, invalidRating, and rating to a separate `renderRatings` functions. it just feels like a bunch of args to have to pass to a function to render a little jsx. also, if we had multiple situations like this in one component, it could start to get confusing as we pass props around all over the place to different functions. 

i'm pretty on the fence about this. for some reason, passing dispatch around as an argument doesn't sit well with me, but i'm not sure thats a valid concern. 

do you still think the separate function approach is better?

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: oh wait, does that mean we want to allow people to write more than one review for an album?

rosscreighton:
> way to render default ComposeReview after successful review is submitted so that it's not locked at the Thank you message

we can accomplish this _super_ easily with redux, i just want to talk to you about the thinking behind the experience before we do it because i'm not entirely sure that i agree.

rosscreighton:
thanks for the review sir 💯

PR #48, opened 2016-08-29
tests for ComposeReview integration

rosscreighton:


imports/ui/components/ComposeReview/ComposeReview.jsx
```diff
@@ -108,4 +112,8 @@ export default connect(state => ({
   invalidRating: state.ComposeReview.invalidRating,
   rating: state.ComposeReview.rating,
   submitSuccess: state.ComposeReview.submitSuccess,
+}), dispatch => ({
```
sameckmeier: cool so im happy youre into using dispatch to props 👍

imports/ui/components/ComposeReview/ComposeReview.tests.js
```diff
@@ -0,0 +1,107 @@
...
+import React from 'react';
+import { chai } from 'meteor/practicalmeteor:chai';
+import { shallow } from 'enzyme';
+import { spy, stub } from 'sinon';
+import { initJSDom } from '../../../lib/testHelpers';
+import { ComposeReview } from './ComposeReview';
+
+initJSDom();
+
+const setup = () => {
```
sameckmeier: i like this as a paradigm 👍

imports/ui/components/ComposeReview/ComposeReview.tests.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const contentEventStub = { target: { value: 'some review content' } };
+
+describe('<ComposeReview />', function () {
+  it('correctly renders selected rating', function () {
+    const { wrapper, props } = setup();
+    const selectedRatingButton = wrapper
+      .find('.ComposeReview-ratings button')
+      .at(props.rating - 1);
```
sameckmeier: we might want to set this to a const named something like selectedRatingIndex

imports/ui/components/ComposeReview/ComposeReview.tests.js
```diff
@@ -0,0 +1,107 @@
...
+
+    chai.expect(selectedRatingButton.hasClass('ComposeReview-selected')).to.equal(true);
+  });
+
+  it('calls setContent when content is changed', function () {
+    const { wrapper, props } = setup();
+
+    chai.expect(props.setContent.called).to.equal(false);
```
sameckmeier: do we need these? like i understand why we might want them, but calledOnce makes sure that the watched function is called only once, so if they were called when they shouldnt calledOnce would return false.

sameckmeier:
looks good!

blocking comments:
making sure that spied functions havent been called before simulating/calling them

imports/ui/components/ComposeReview/ComposeReview.jsx
rosscreighton: I definitely am, but that wasn't what prompted this change. 

If you don't use map dispatch to props, you have to stub es6 imports in your tests. Stubbing props is MUCH easier than stubbing file imports.

imports/ui/components/ComposeReview/ComposeReview.tests.js
rosscreighton: Props to redux docs

imports/ui/components/ComposeReview/ComposeReview.tests.js
rosscreighton: Good call

imports/ui/components/ComposeReview/ComposeReview.tests.js
rosscreighton: I see your point, but my thinking was that this ensures we don't get a false positive. Let's say the component mounts and  the setContent prop is called for some reason when it shouldn't be, then we change the text in the textarea and setContent is not called like it should be. Now we have a false positive.

I'm definitely open to the idea that I'm over engineering this though

rosscreighton:
@sameckmeier  addressed your comments, merging

PR #50, opened 2016-08-30
Modal

sameckmeier:
ill remove that modal button rendered inside our Page component -- just wanted to leave it for proof of concept. 

main question is whether we want to set break points for our modal or leave it as is?


imports/redux/ducks/Modal.js
```diff
@@ -0,0 +1,27 @@
+const SET_DISPLAY_MODAL = '/gm/modal/set-display-modal';
+const SET_HIDE_MODAL = '/gm/modal/set-hide-modal';
+
+export const setDiplayModal = () => ({
+  type: SET_DISPLAY_MODAL,
+  display: true,
```
rosscreighton: i don't think this action needs any data other than a type. `display: true` and `display: false` are implicit to the `SET_DISPLAY_MODAL` and `SET_HIDE_MODAL` actions. in the reducer, we don't need to do `{ display: action.display }`, we could just say `{ display: true }` or `{ display: false }` based on the action type.

imports/redux/ducks/Modal.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+export const setHideModal = () => ({
```
rosscreighton: i have a naming suggestion.

by definition, redux actions describe an event that occurs in _the app_. it seems your action and action creator names are describing an event that occurs _in the reducer_(setting display to true or false).

i think it makes more sense to omit the word 'set' from these names. 

when i am in some component and i want to launch a modal, i think it makes more sense to say `dispatch(displayModal())` instead of `dispatch(setDisplayModal())`. the component isn't "setting" anything, it just wants to launch a modal.

imports/redux/reducers.js
```diff
@@ -1,4 +1,5 @@
 import { combineReducers } from 'redux';
+import Modal from './ducks/Modal';
```
rosscreighton: we'll have to wait to merge this until all the compose review stuff is on master because i completely re-worked the redux piece of the app

rosscreighton:
❗ tests???

imports/ui/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function Modal({
+  dispatch,
+  display,
+  children,
+ }) {
+  const displayToggle = display ? 'Modal-display' : 'Modal-hide';
+  return (
+    <div className={`Modal ${displayToggle}`}>
+      <div className='Modal-overlay' onClick={() => dispatch(setHideModal())}/>
```
rosscreighton: should we also listen for the the `esc` keypress to close the modal?

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
+.Modal {
+  $box-size: 40px;
+
+  &.Modal-display {
+    display: block;
```
rosscreighton: do you think it makes sense to show/hide the modal using css?

we could save on initial render time by NOT rendering the modal markup until DISPLAY_MODAL is dispatched. then, when HIDE_MODAL was dispatched, the markup would be removed from the DOM by react.

part of me thinks this doesn't really matter, but another part is reminded that rendering performance is a WAY bigger concern on mobile. things like this that have no effect on desktop can be noticeably slow on mobile.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  &.Modal-hide {
+    display: none;
+  }
+
+  .Modal-overlay, .Modal-modal {
```
rosscreighton: this positioning doesn't seem right for .Modal-modal. this indicates to me that the modal takes up the entire screen.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+    font-size: $font-size-lg;
+    border: 1px solid $color-gray;
+    background-color: $color-baby-powder;
+    height: $box-size;
+    width: $box-size;
+  }
+
+  .Modal-overlay {
+    z-index: 9999998;
+    width: 100%;
```
rosscreighton: does the overlay need a width and height?

i think the following from above should make the overlay take up the entire screen without having to set a width and height:

position: absolute;
top: 0;
bottom: 0;
left: 0;
right: 0;

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+    height: 100%;
+    background-color: $color-gray;
+    opacity: 0.6;
+  }
+
+  .Modal-modal {
+    z-index: 9999999;
```
rosscreighton: we should put all z-index values in a variables file so if we use z-index in the future, we can reference the different z-levels that exist in the project

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+    width: 50%;
+    height: 50%;
```
rosscreighton: should the modal have a fixed width/height, or should it expand/contract to fit its contents?

i think it would be more flexible if its width/height was dynamic. this way we could render a component inside the modal and the component could have a complex layout with a bunch of media queries but the modal would still look right not matter what size its contents changed to.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
+.Modal {
```
rosscreighton: this file was very organized, logical, easy to read. 💯

imports/ui/components/Page/Page.jsx
```diff
@@ -1,13 +1,28 @@
...
   return (
     <div>
       <NavBar />
       <div className="Page">
         {children}
       </div>
+      <button onClick={() => dispatch(setDiplayModal())}>MODAL</button>
+      <Modal>
+        <div>
+          <p>SWEEEEEEET</p>
```
rosscreighton: what happens if we want to have multiple possible modals on a page?
what if we want the contents of modal to change based on some value in the store?

the way this abstraction is currently set up, we can only have _one_ modal on a page and it's contents cannot be changed.

to make the abstraction more robust, i would suggest storing the contents of the modal in the redux store under `state.Modal.content`. something like this:

```
dispatch(displayModal(<SomeComponent />))
```

in the reducer:

```
case SET_DISPLAY_MODAL:
      return Object.assign({}, state, { display: true, content: action.content })
```

then, in the Modal component:

```

function Modal({ display, content}) {
  return display ? content : null;
}

  display: state.Modal.display,
  content: state.Modal.content,
}))(Modal);
```

rosscreighton:
most of the comments above are blocking. also need to write a few tests for this 😎.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+    bottom: 0;
+    left: 0;
+    right: 0;
+    margin: auto;
+  }
+
+  .Modal-close {
+    font-weight: bold;
+    font-size: $font-size-lg;
+    border: 1px solid $color-gray;
```
rosscreighton: this is probably just personal preference but i'm not a big fan of the border around the X.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  .Modal-overlay {
+    z-index: 9999998;
+    width: 100%;
+    height: 100%;
+    background-color: $color-gray;
+    opacity: 0.6;
+  }
+
+  .Modal-modal {
```
rosscreighton: design trend wise, i think modals usually have a `box-shadow` to help them pop a little bit. i might consider adding a very slight box shadow here.

rosscreighton:
oh and to answer your question re: breakpoints. i think we definitely need to a mobile breakpoint. i think when a modal pops on mobile, it should take up the entire screen, but when it pops on desktop it should have a max-width like 80% or something.

imports/redux/ducks/Modal.js
sameckmeier: very good point! changing this

imports/redux/ducks/Modal.js
sameckmeier: mmmm makes sense. i like this a paradigm: if an action creator is supposed to set a bool val then we do so in the reducer and have a descriptive type.

imports/ui/components/Modal/Modal.jsx
sameckmeier: i was trying to set that up last night but was having some troubles figuring it out. definatley something im gonna research more

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -0,0 +1,47 @@
...
+    display: none;
+  }
+
+  .Modal-overlay, .Modal-modal {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    margin: auto;
```
sameckmeier: it centers it, but there could be a better way.

imports/ui/stylesheets/components/Modal.scss
sameckmeier: if we can save render time by going this route, then im down. didnt even consider this!

sameckmeier:
im conflicted over how i setup the esc keydown event. there was another way i got it to work but there were some realistic scenarios that might create for funky behavior. would love to hear your thoughts on it. 

additionally, the esc keydown event caused me to change Modal from being functional to an es6 component. 

finally, there are 2 tests that ive commented out because im still trying to figure out how to test one of them and the other just doesnt work due to enzyme. i threw discussion points up in trello for em.

.eslintrc
```diff
@@ -12,3 +12,9 @@ globals:
 rules:
   import/no-unresolved: off
   no-underscore-dangle: off
+  max-len: [2, 120, 2]
```
rosscreighton: did you make these changes manually? should just be able to merge in master.

imports/ui/components/Page/Page.jsx
```diff
@@ -2,27 +2,31 @@ import React from 'react';
...
+          </div>
+        </Modal>
+      )}
     </div>
   );
 }
 
-export default connect()(Page);
+export default connect(state => ({
+  display: state.Modal.display,
```
rosscreighton: i would call this prop displayModal since we are in the context of the Page component. would be less confusing if the reader knew that the prop was related to the modal.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -28,15 +11,19 @@
   }
 
   .Modal-overlay {
```
rosscreighton: excellent

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -24,11 +27,22 @@
...
-    height: 50%;
+    min-width: $modal-size-sm;
+    min-height: $modal-size-sm;
     padding: $gutter-sm;
     border-radius: $border-radius;
     border: 1px solid $color-gray;
     background-color: $color-baby-powder;
+    box-shadow: 0px 0px 15px $color-black;
+
+    @media screen and (min-width: $breakpoint-tablet) {
```
rosscreighton: i wouldn't use min-width for the modal. i think we want to use max-width. the modal would re-size to fit any content that is inside of it, _up to_ certain maximum dimensions. this way the modal would always look good on any screen size.

so technically,if there were no content in the modal, the modal would have zero height and zero width.

imports/ui/stylesheets/z-index.scss
```diff
@@ -0,0 +1 @@
+$z-index-modal: 9999999;
```
rosscreighton: nice

imports/ui/components/Modal/Modal.jsx
```diff
@@ -1,30 +1,45 @@
...
 Modal.propTypes = {
   display: PropTypes.bool.isRequired,
+  hideModalOnClick: PropTypes.func.isRequired,
+  hideModalOnKeyDown:PropTypes.func.isRequired,
 };
 
 export default connect(state => ({
   display: state.Modal.display,
+}), dispatch => ({
+  hideModalOnClick: (e) => { dispatch(hideModal()) },
```
rosscreighton: i would name these props the same name as the action creator.

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -13,20 +21,18 @@
...
 
-  .Modal-overlay {
-    top: 0;
-    right: 0;
-    left: 0;
-    bottom: 0;
-    z-index: $z-index-modal;
-    position: absolute;
-    background-color: rgba(73, 73, 73, 0.6);
+  .Modal-wrapper {
```
rosscreighton: why do we need this wrapper class?
if we do need a wrapper class, why not just use the outer component div: `.Modal`?

imports/lib/keyCodes.js
```diff
@@ -0,0 +1 @@
+export const ESC = 27;
```
rosscreighton: good idea

imports/ui/stylesheets/z-index.scss
rosscreighton: should we have a hyphen in this file name? we usually camelCase file names.

imports/redux/ducks/Modal.js
```diff
@@ -9,6 +10,11 @@ export const hideModal = () => ({
   type: HIDE_MODAL,
 });
 
+export const setModalContent = (content) => ({
```
rosscreighton: i don't think we need a separate action for setting the modal content. when we display the modal, we could just pass along the content we want to display in it: 

```
export const diplayModal = () => ({
    type: DISPLAY_MODAL,
    content: some text or a react component
})

File: imports/ui/components/Modal/Modal.jsx [hunk 24]
```diff
@@ -32,14 +33,14 @@ export class Modal extends Component {
...
   hideModalOnClick: PropTypes.func.isRequired,
   hideModalOnKeyDown:PropTypes.func.isRequired,
 };
 
 export default connect(state => ({
-  display: state.Modal.display,
+  content: state.Modal.content,
 }), dispatch => ({
   hideModalOnClick: (e) => { dispatch(hideModal()) },
-  hideModalOnKeyDown: (e) => { if (e.keyCode == 27) { dispatch(hideModal()) } },
```
rosscreighton: i suggested a name change for these earlier, but now i see why you named these this way. i retract my re-naming suggestion. this makes sense

imports/ui/components/Modal/Modal.jsx
```diff
@@ -32,14 +33,14 @@ export class Modal extends Component {
...
+  hideModalOnKeyDown: (e) => { if (e.keyCode == ESC) { dispatch(hideModal()) } },
```
rosscreighton: in javascript it's good practice to avoid `==` whenever possible. i might re-work this to use `===`, and cast `e.keycode` to a Number before comparison if needed.

imports/ui/components/Page/Page.jsx
```diff
@@ -16,19 +16,23 @@ export function Page({
...
+  </div>
+);
+
 export default connect(state => ({
   display: state.Modal.display,
 }), dispatch => ({
-  diplayModal: () => dispatch(diplayModal()),
+  diplayModal: () => {
+    dispatch(setModalContent(ModalContent));
+    dispatch(diplayModal());
```
rosscreighton: again, i would just set `content` on the displayModal action so we don't have to have two actions

imports/ui/components/Modal/Modal.jsx
```diff
@@ -41,6 +41,6 @@ Modal.propTypes = {
 export default connect(state => ({
   content: state.Modal.content,
 }), dispatch => ({
-  hideModalOnClick: (e) => { dispatch(hideModal()) },
-  hideModalOnKeyDown: (e) => { if (e.keyCode == ESC) { dispatch(hideModal()) } },
+  hideModalOnClick: () => { dispatch(hideModal()) },
+  hideModalOnKeyDown: (e) => { e.stopPropagation(); if (e.keyCode == ESC) { dispatch(hideModal()) } },
```
rosscreighton: do we want to stopPropagation for _every_ keypress event, or just for ESC keypress events

imports/redux/ducks/Modal.tests.js
```diff
@@ -0,0 +1,37 @@
...
+import reducer, {
+  defaultState,
+  DISPLAY_MODAL,
+  HIDE_MODAL,
+  SET_MODAL_CONTENT,
+} from './Modal';
+import React from 'react';
+import { chai } from 'meteor/practicalmeteor:chai';
+
+describe('the Modal reducer', () => {
```
rosscreighton: this test looks good 👍

imports/ui/components/Modal/Modal.tests.js
```diff
@@ -0,0 +1,85 @@
...
+
+  return { wrapper, props };
+};
+
+describe('<Modal />', function () {
+  it('should add keydown listener to the document', function () {
+    let calledAddEventListener = false;
+
+    const addEventListener = () => calledAddEventListener = true;
+    const stubbedAddEventListener = sinon.stub(document, 'addEventListener', addEventListener);
```
rosscreighton: can we `import { stub } from 'sinon'` and call `stub()` instead of `sinon.stub()`?

imports/ui/components/Modal/Modal.tests.js
```diff
@@ -0,0 +1,85 @@
...
+
+  it('should render content inside Modal', function () {
+    const { wrapper, props } = setup();
+
+    const content = wrapper.find('.Modal-content').children().first();
+
+    chai.expect(content.hasClass('test')).to.equal(true);
+  });
+
+  // need to figure out how to test this
```
rosscreighton: yeah, i wonder if there's a way to simulate a keypress

imports/ui/components/Modal/Modal.tests.js
```diff
@@ -0,0 +1,85 @@
...
+  // it('should hide Modal on ESC keydown', function () {
+  //
+  // });
+
+  // If React is imported before we initialize jsdom, then we run into the issue of unmount not working https://github.com/airbnb/enzyme/issues/395
```
rosscreighton: hmmmmm. lets talk about this on the phone

rosscreighton:
looks good dude. i made a bunch of comments. if a comment is blocking i noted that in the comment.

imports/ui/components/Modal/Modal.tests.js
sameckmeier: we can indeed 👍 need to ingrain in my head that we want to import functions over objs in most cases

imports/ui/stylesheets/components/Modal.scss
sameckmeier: ohhhhhh gotcha 👍 changing this.

imports/ui/stylesheets/z-index.scss
sameckmeier: im indifferent. like we're using hyphens for both classnames and filenames, but if we want to camelCase everything, id be into it for consistency sake.

imports/ui/components/Page/Page.jsx
sameckmeier: would that be redundant though? like im always down more more explicit naming, but only asking since we're calling display on the Modal odj.

imports/ui/stylesheets/components/Modal.scss
sameckmeier: mmmm good catch

imports/redux/ducks/Modal.js
sameckmeier: yup another good catch 👍

imports/ui/components/Modal/Modal.jsx
sameckmeier: 👍

imports/ui/components/Modal/Modal.jsx
sameckmeier: ahhhh misunderstanding on stopping propagation -- first it hits it targeted element and if it's an ESC then stopPropagation

sameckmeier:
alrighty, gonna push these changes, and change Modal back to a functional component.

imports/ui/components/Page/Page.jsx
rosscreighton: my thinking was that we're passing the prop into the Page component, not the Modal component, so we want to make sure the reader doesn't think that prop means something like "display the page". if we add "modal" to the prop name, the reader would know that that prop means "display the modal".

i don't feel super strongly about this, cool with me if you leave it as is.

imports/ui/stylesheets/z-index.scss
rosscreighton: i think i want to consistency, if you're down.

imports/ui/components/Modal/Modal.jsx
rosscreighton: yeah that makes sense.

we actually probably don't really need to stopPropagation for an ESC key press though. usually you want to stopPropagation or preventDefault on a form submission event or something that the browser would take a default action on.

imports/redux/ducks/Modal.js
```diff
@@ -1,14 +1,28 @@
...
+import { hideModalOnKeydown } from '../../ui/components/Modal/utils';
 
-export const diplayModal = (content) => ({
-  type: DISPLAY_MODAL,
+export const SETUP_MODAL = '/gm/modal/display-modal';
+export const TEARDOWN_MODAL = '/gm/modal/hide-modal';
+
+export const diplayModal = content => (
+  dispatch => {
+    document.addEventListener('keydown', hideModalOnKeydown.bind(null, dispatch));
```
rosscreighton: nice use of bind

rosscreighton: fucking js ninja over here

imports/redux/ducks/Modal.js
```diff
@@ -1,14 +1,28 @@
-export const DISPLAY_MODAL = '/gm/modal/display-modal';
-export const HIDE_MODAL = '/gm/modal/hide-modal';
+import { hideModalOnKeydown } from '../../ui/components/Modal/utils';
```
rosscreighton: it feels to me like this function really belongs in the ducks file. our components directory really only contains functions accept props and render jsx - this function manages state, so i feel like it belongs in redux land.

imports/ui/components/Modal/utils.js
```diff
@@ -0,0 +1,9 @@
+import { ESC } from '../../../lib/keyCodes';
+import { hideModal } from '../../../redux/ducks/Modal';
```
rosscreighton: yeah i would just put this function in the ducks file. the fact that each file is importing something from the other indicates to me that they should just be in the same file.

imports/ui/components/Modal/Modal.jsx
```diff
@@ -1,44 +1,31 @@
...
-  hideModalOnKeyDown:PropTypes.func.isRequired,
+  hideModal: PropTypes.func.isRequired,
 };
 
 export default connect(state => ({
   content: state.Modal.content,
 }), dispatch => ({
-  hideModalOnClick: () => { dispatch(hideModal()) },
-  hideModalOnKeyDown: (e) => { if (Number(e.keyCode) === ESC) { e.stopPropagation(); dispatch(hideModal()); } },
+  hideModal: () => { dispatch(hideModal()) },
```
rosscreighton: you can omit curly braces around the function body here

imports/ui/components/Modal/Modal.tests.js
```diff
@@ -16,45 +16,31 @@ const setup = () => {
 
   const props = {
     content,
-    hideModalOnKeyDown: spy(),
-    hideModalOnClick: spy(),
+    hideModal: spy(),
   };
 
-  const wrapper = mount(<Modal {...props} />);
```
rosscreighton: fuck yeah for being able to not use mount

imports/ui/components/Modal/Modal.tests.js
```diff
@@ -16,45 +16,31 @@ const setup = () => {
...
+  const wrapper = shallow(<Modal {...props} />);
 
   return { wrapper, props };
 };
 
 describe('<Modal />', function () {
-  it('should add keydown listener to the document', function () {
```
rosscreighton: this functionality was moved to an action creator and is therefore no longer tested.

should we talk more seriously about testing action creators?

rosscreighton:
looks good!

nothing blocking.

would be nice to have:
- move hideModalOnKeydown to /ducks/Modal.js
- test coverage for the adding and removing of event listeners

:shipit:

imports/redux/ducks/Modal.js
```diff
@@ -17,10 +31,10 @@ export const defaultState = {
 
 export default function reducer(state = defaultState, action) {
   switch (action.type) {
-    case DISPLAY_MODAL:
+    case SETUP_MODAL:
       return Object.assign({}, state, { display: true, content: action.content });
-    case HIDE_MODAL:
-      return Object.assign({}, state, { display: false });
+    case TEARDOWN_MODAL:
+      return Object.assign({}, state, defaultState);
```
sameckmeier: i was thinking that, but i believe redux suggests that one should return a new obj when possible

rosscreighton:
we should try again to get those tests working before we merge this

imports/redux/ducks/Modal.js
sameckmeier: ya i was town on this because it didnt fit perfectly in the ducks module, but does alter state. i threw the utils in there just as like an alternative, but also prefer ducks to this, so ill change back before pushing 👍

imports/ui/components/Modal/Modal.jsx
sameckmeier: ah right good catch 👍

imports/ui/components/Modal/Modal.tests.js
sameckmeier: 💥

imports/ui/components/Modal/Modal.tests.js
sameckmeier: i was thinking as we use more side effects/thunk methods, it makes sense to throw some tests up, but i still am on the fence as to whether they're a huge priority? either, im down to chat more on it.

sameckmeier:
so for those event listeners, it'd be tough to test them outright, so i was thinking of creating a class that encapsulates add and remove eventlisteners and would have two functions, create and destroy, and we could spy em and make sure they were called. or, we can just begin testing action creators?

imports/redux/ducks/Modal.js
rosscreighton: ah that makes sense.

rosscreighton:
i think i would prefer to test action creators, but i'm not sure i entirely understand your class idea, so maybe lets talk about this on the phone?

sameckmeier:
something along the lines of this: http://www.jstips.co/en/DOM-event-listening-made-easy/

sameckmeier:
i actually really like the idea of a function returning a handle and using that to remove events, but we could also do a class that accepts a few args and we'd do something like eventHandler.add and eventHandler.remove. either way, we'd just make sure that add and remove event was called via spy.

imports/redux/ducks/Modal.tests.js
```diff
@@ -31,3 +34,20 @@ describe('the Modal reducer', () => {
     });
   });
 });
+
+describe('Modal action creators', () => {
```
rosscreighton: 👍 glad we decided to test these

imports/redux/ducks/Modal.tests.js
```diff
@@ -31,3 +34,20 @@ describe('the Modal reducer', () => {
...
+  const dispatch = spy();
+
+  it('displayModal', () => {
```
rosscreighton: can we give these cases more descriptive names? like "displayModal adds event listener to document"

imports/redux/ducks/Modal.tests.js
```diff
@@ -12,42 +12,43 @@ import React from 'react';
 import { chai } from 'meteor/practicalmeteor:chai';
 import { spy, stub } from 'sinon';
 
-describe('the Modal reducer', () => {
```
rosscreighton: good call

rosscreighton:
:shipit:

imports/redux/ducks/Modal.tests.js
sameckmeier: we absolutely can and should

PR #51, opened 2016-08-31
timestamps

rosscreighton:


imports/ui/components/Review/Review.jsx
```diff
@@ -1,10 +1,11 @@
 import React, { PropTypes } from 'react';
+import moment from 'moment';
```
rosscreighton: it sucks we have to import the entire library just to use one function.

rosscreighton: js bundle size really does make a big difference on mobile, so we should try to avoid stuff like this when we can

imports/api/AlbumFollows/AlbumFollows.js
```diff
@@ -7,5 +7,4 @@ export const schema = new SimpleSchema({
   followerId: { type: Number },
 });
 
-AlbumFollows.schema = schema;
-AlbumFollows.attachSchema(AlbumFollows.schema);
+AlbumFollows.attachSchema(schema);
```
sameckmeier: 👍 most excellent

imports/api/reviews/Reviews.js
```diff
@@ -9,4 +9,11 @@ export const schema = new SimpleSchema({
   content: { type: String },
 });
 
+Reviews.before.insert((userId, doc) => {
+  if (Meteor.isServer) {
+    // eslint-disable-next-line no-param-reassign
+    doc.createdAt = new Date();
```
sameckmeier: do we need to have a corresponding createdAt key defined on schema?

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -96,7 +96,7 @@ function AlbumPageLoader({
 
     if (albumHandle.ready()) {
       const album = Albums.findOne({ slug });
-      const reviews = Reviews.find({ albumId: album._id }).fetch();
+      const reviews = Reviews.find({ albumId: album._id }, { sort: { createdAt: -1 } }).fetch();
```
sameckmeier: nice

imports/lib/collectionHelpers.js
```diff
@@ -0,0 +1,8 @@
+export const timestamps = collection => {
```
sameckmeier: 👍

imports/ui/components/Review/Review.jsx
```diff
@@ -4,7 +4,7 @@ import TruncatedText from '../TruncatedText';
 export default function Review({
   authorName,
   authorAvatarUrl,
-  timeAgo,
+  createdAt,
```
sameckmeier: 👍

imports/lib/collectionHelpers.js
sameckmeier: alternatively, could we create a collection callbacks file, import necessary collections, iterate over em and set up these callbacks? just a thought, especially if we add more hooks.

sameckmeier:
only blocking comments: using moment and possibly changing out how and where we're implementing those hooks. 

other than that, looks good!

imports/ui/components/Review/Review.jsx
rosscreighton: yeah i was thinking about doing that. i'll try it and see how shisty it turns out to be

imports/lib/collectionHelpers.js
rosscreighton: definitely think we should do something like that if we add more callbacks

rosscreighton: i'm down to do it now too though

imports/api/reviews/Reviews.js
rosscreighton: yes, but we can't for reasons that i don't want to type out here. i'm adding something to trello to explain to you on the phone.

sameckmeier:
:shipit:

PR #53, opened 2016-09-04
modal size

rosscreighton:


rosscreighton:
there is some login ui scaffolding here that would be a bitch to remove so i'm leaving it :/

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -24,32 +20,55 @@
...
   .Modal-modal {
     z-index: $z-index-modal;
-    max-width: $modal-size-sm;
-    max-height: $modal-size-sm;
-    padding: $gutter-sm;
     border-radius: $border-radius;
     border: 1px solid $color-gray;
     background-color: $color-baby-powder;
-    box-shadow: 0px 0px 15px $color-black;
+    width: 95%;
```
sameckmeier: do we need none? can we just push all of this classes styles into Modal-modal?

imports/ui/stylesheets/components/Modal.scss
```diff
@@ -24,32 +20,55 @@
...
+    height: 95%;
 
-    @media screen and (min-width: $breakpoint-tablet) {
-      max-width: $modal-size-md;
-      max-height: $modal-size-md;
+    &.Modal-size-none {
```
sameckmeier: can we merge this class into Modal-modal? we could then make set height 95% in size-sm/lg. might just make things a but more explicit.

sameckmeier:
nothing blocking 👍 just a comment on Modal-size-none. let's :shipit: if you dont think the changes i proposed are worth it

imports/ui/stylesheets/components/Modal.scss
rosscreighton: @sameckmeier ah yeah thats a good point, but i think the problem here is naming. the word "none" doesn't actually mean there is no size, it means "the modal is sized dynamically based on its contents". we do need a separate class to manage the styles for the dynamic behavior, but i think it needs a better name. maybe Modal-size-dynamic?

rosscreighton:
just realized there is a test failing on this branch and the test runner is not giving me the failure details 😑😑😑

rosscreighton:
@sameckmeier got the tests passing, merging this.

PR #54, opened 2016-09-06
Error handling

sameckmeier:
im not stoked with how we're testing these thunk action creators: we're making sure tht the correct number of dispatches are executed. as it stands, it's difficult to spy on the actions creators that we're dispatching; we'd either have to inject them as an arg or export our modules as objs, neither of which sound appealing. Then again, all that we want to ensure is that these thunk functions are dispatching, which is their core functionality. If we were to test the specific action creators that we're dispatching, then we might be coupling the tests to our code too tightly.


imports/lib/errorUtils.js
```diff
@@ -0,0 +1,10 @@
+export const INVALID_REVIEW_CLIENT = {
+  error: 'ComposeReview.submitReview.invalidReview',
+  reason: 'user did not provide a rating and/or content',
+  details: 'Looks like you forgot to select a rating or write a review'
+};
+
+export const generateError = (args) => {
+  const { error, reason, details } = args;
+  return new Meteor.Error(error, reason, details);
```
rosscreighton: are we sure we want this to be a Meteor.Error? do we want to have our client side error reporting depend on the meteor framework?

imports/redux/ducks/ErrorMessage.js
```diff
@@ -0,0 +1,48 @@
...
+
+export const setModalContentToDisplay = modalContentToDisplay => ({
+  type: SET_MODAL_CONTENT_TO_DISPLAY,
+  modalContentToDisplay,
+});
+
+export const handleError = err => (dispatch, getState) => {
+  const currentState = getState();
+  const { content } = currentState.Modal;
+  const { details } = err;
```
rosscreighton: do all errors have this details attribute?

rosscreighton:
will take me a little while to review this one today

rosscreighton:
so i think we could simplify this a lot. right now the ErrorMessage reducer is doing a lot of things: pulling shit from the Modal state, dispatching modals, caching previous modal content, handling a duck-typed error object.

rosscreighton:
i propose handling errors like so, using the Page component:

```
import { connect } from 'react-redux';
import NavBar from '../NavBar';
import Modal from '../Modal';
import ErrorMessage from '../ErrorMessage';

export function Page({
  children,
  error,
  displayModal,
 }) {
  return (
    <div>
      <NavBar />
      <div className="Page">
        {children}
      </div>
      {error && <Modal><ErrorMessage error={error} /></Modal>}
      {displayModal && !error && <Modal />}
    </div>
  );
}

Page.propTypes = {
  displayModal: PropTypes.bool.isRequired,
  children: PropTypes.element,
};

  displayModal: state.Modal.display,
  error: state.Error.error,
}))(Page);
```

this way, the _only_ thing that the Error reducer has to do is add or a remove an error from the state, it doesn't have to cache previous modal content, inquire upon the Modal state, dispatch modals, none of that. 

if an error is present in state, that error will override any modal content that is also in the state. if there is an error, we will always display it in the modal. once the error is removed from redux, we can go back to displaying whatever modal content is the Modal state.

rosscreighton:
@sameckmeier why did we merge this one? error-handling-alt PR was against master, which we merged, but then we also merged this one, which has all the old error handling logic in it.

sameckmeier:
ah shit, i thought this alt

sameckmeier:
though i dont think i actually merged this. was thinking about it and it's possible that because i cloned error handling into alt, and closed alt, git may have recognized this branch as being merged once i merged error handling alt?

PR #55, opened 2016-09-09
Error handling alt

rosscreighton:


imports/redux/ducks/Error.js
```diff
@@ -0,0 +1,44 @@
...
+
+  dispatch(setError(null));
+};
+
+export const defaultState = { error: null };
+
+export default function reducer(state = defaultState, action) {
+  switch (action.type) {
+    case SET_ERROR:
+      return Object.assign({}, state, { error: action.error });
```
rosscreighton: we can just use `{ ...action.error }` since we don't need to preserve any keys on state. the only key on Error state is `error`, so we don't need to merge new state into old state, when just provide an entirely new state with the error key.

imports/redux/ducks/Error.tests.js
```diff
@@ -0,0 +1,116 @@
...
+  describe('action creators', () => {
+    describe('handleError', () => {
+      describe('when state.Modal.display is false', () => {
+        it('should dispatch displayModal', () => {
+          const { dispatch, getState, error } = setup();
+          getState.returns({ Modal: { display: false } });
+
+          handleError(error)(dispatch, getState);
+
+          chai.expect(dispatch.calledTwice).to.equal(true);
```
rosscreighton: this doesn't feel awesome but i think its fine. i took a quick look at the sinon docs and couldn't really think of a better way to test this.

rosscreighton: i take that back, i think we should use `dispatch.calledWith( _whatever the action creator returns_ )` to ensure that the correct things are being dispatched.

imports/redux/ducks/Modal.js
```diff
@@ -2,27 +2,31 @@
...
   }
 );
 
-export const teardownModal = () => ({ type: TEARDOWN_MODAL });
-
 export const hideModal = () => (
   dispatch => {
     document.removeEventListener('keydown', hideModalOnEsc);
-    dispatch(teardownModal());
+    dispatch(setDisplay(false));
```
rosscreighton: shouldn't `hideModal` also set `{ content: null }`?

imports/redux/ducks/Modal.js
```diff
@@ -2,27 +2,31 @@
...
+export const setDisplay = display => ({
+  type: SET_DISPLAY,
+  display,
+});
+
 export const displayModal = content => (
   dispatch => {
     document.addEventListener('keydown', hideModalOnEsc.bind(null, dispatch));
-    dispatch(setupModal(content));
+    if (content !== undefined) { dispatch(setContent(content)) }
```
rosscreighton: should we put a comment here explaining that this is actually for showing an error?

imports/redux/ducks/Modal.tests.js
```diff
@@ -13,19 +13,70 @@ import { chai } from 'meteor/practicalmeteor:chai';
...
+        const { content, dispatch } = setup();
+        const addEventListener = spy(document, 'addEventListener');
+
+        displayModal(content)(dispatch);
+        chai.expect(addEventListener.calledOnce);
+      });
+    });
+
+    describe('hideModal', () => {
+      it('should dispatch setDisplay', () => {
```
rosscreighton: should also dispatch `setContent(null)`?

imports/redux/ducks/Modal.tests.js
```diff
@@ -13,19 +13,70 @@ import { chai } from 'meteor/practicalmeteor:chai';
...
-  const content = (<div>This is a test</div>);
+  const setup = () => {
+    const content = (<div>This is a test</div>);
+    const display = true;
+    const dispatch = spy();
+
+    return { content, display, dispatch };
+  };
+
+  describe('action creators', () => {
```
rosscreighton: again, i want to consider using `spy.calledWith(args)` in these tests

imports/ui/components/ErrorMessage/ErrorMessage.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { PropTypes } from 'react';
```
rosscreighton: this component is beautiful

imports/ui/components/ErrorMessage/ErrorMessage.tests.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-env mocha */
+/* eslint-disable func-names, prefer-arrow-callback */
+
+import React from 'react';
+import { ErrorMessage } from './ErrorMessage';
+import { chai } from 'meteor/practicalmeteor:chai';
```
rosscreighton: lets import this from the npm package

imports/ui/components/ErrorMessage/ErrorMessage.tests.js
```diff
@@ -0,0 +1,40 @@
...
+    const { wrapper, props } = setup();
+    chai.expect(wrapper.find('p').text()).to.equal(props.details);
+  });
+
+  it('calls closeErrorMessage on button click', () => {
+    const { wrapper, props } = setup();
+    wrapper.find('button').simulate('click');
+    chai.expect(props.closeErrorMessage.calledOnce).to.equal(true);
+  });
+});
```
rosscreighton: these ErrorMessage tests are perfect

imports/ui/components/Modal/Modal.jsx
```diff
@@ -12,20 +13,22 @@ export function Modal({
       <div className='Modal-modal'>
         <button className='Modal-close' onClick={() => hideModal()}>X</button>
         <div className='Modal-content'>
-          {content}
+          {error || content}
```
rosscreighton: excellent. this was super easy.

i still don't think this is THEE best possible solution to our problem, but i feel good about the solution and it definitely more that suites our needs for now.

imports/ui/stylesheets/components/ErrorMessage.scss
```diff
@@ -0,0 +1,14 @@
+.ErrorMessage {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+
+  button {
```
rosscreighton: these styles are repeated from ComposeReivew. we should move these styles to a utility class in global.scss or utilities.scss.

i'm thinking it could be something like `.gm-button`, and we could create the look and feel for our buttons using this class and just apply it when needed.

rosscreighton:
good shit dude, i think this solves our problem 👍 

i indicated which comments are blocking.

imports/redux/ducks/Error.js
sameckmeier: right, so we could just return { error: action.error }? { ...action.error} would throw an error because it's not a hash? just wondering because i might misunderstand the spread operator in this use case

imports/redux/ducks/Modal.js
sameckmeier: you are absolutely correct...brain farts on my part

imports/redux/ducks/Error.js
rosscreighton: Oh yeah you're right error is not a hash so you wouldn't need the spread operator.

imports/redux/ducks/Error.tests.js
sameckmeier: fuck man, good find! i didnt even see it in the docs! ive been messing around with it and like the use of chai.expect(dispatch.calledWith(match.object)).to.equal(true) for example, ie, just making sure that dispatch is being called with the proper arg type so many times. i feel like any more specificity might couple our tests too tightly.  what are your thoughts?

imports/redux/ducks/Modal.js
sameckmeier: 👍

imports/redux/ducks/Error.tests.js
rosscreighton: so i think that checking that dispatch is called with the right arg type is basically the same thing as checking that it was called. i don't think that provides any extra info.

if we're going to use calledWith, i'd want to check the actual contents of the action object that passed to dispatch. like just ensure that actionObject.type is the action type string that we expect dispatch to be called with.

imports/redux/ducks/Modal.tests.js
sameckmeier: a similar brain fart lol

imports/ui/components/ErrorMessage/ErrorMessage.tests.js
sameckmeier: oh shit, that's right 👍

imports/ui/components/ErrorMessage/ErrorMessage.tests.js
sameckmeier: 💥

imports/ui/components/ComposeReview/ComposeReview.tests.js
```diff
@@ -52,6 +52,7 @@ describe('<ComposeReview />', function () {
 
   it('calls setContent when content is changed', function (done) {
     const { wrapper, props } = setup();
+    this.timeout(5000);
```
rosscreighton: @sameckmeier this doesn't seem very cool

sameckmeier: agreed. this a temp fix just to make sure all of the tests pass. we need to figure out a way to implement a test config file of some kind.

PR #56, opened 2016-09-12
Side panel

sameckmeier:
side panel layout, other albums, and tracks are all in this one, so kind of a fatty.

we still need to set a mobile version of this and need to turn those other albums into links (we should chat about this for a minute one last time, but my intial thoughts is that, and i believe this is where we ended up on this the last time, is that when one clicks on an album link, it dispatches a createAlbum or navigateToAlbum in which it checks if the album exists, creates it and the artist if not, and then navigates to the page.


imports/redux/ducks/Track.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export default function reducer(state = defaultState, action) {
+  switch (action.type) {
+    case SET_PREVIEW_TRACK_ID:
+      return { previewTrackId: action.previewTrackId };
+    default:
+      return state;
+  }
+}
```
rosscreighton: reading through this file, its not immediately obvious to me what the purpose of this reducer is. i'm sure it will get clearer as I review the rest of the code, but i just wanted to note this because the more self explanatory code is the better

imports/lib/symbolCodes.js
```diff
@@ -0,0 +1 @@
+export const rightBlackPointer = '\u25ba';
```
rosscreighton: do these codes specify a color? like would this symbol always be black or could the color be changed

imports/api/Albums/server/publications.js
```diff
@@ -1,3 +1,4 @@
-import { Albums } from '../Albums';
+import { Albums, ARTIST_ALBUMS_LIMIT } from '../Albums';
 
 Meteor.publish('album', slug => Albums.find({ slug }));
+Meteor.publish('artistAlbums', artistId => Albums.find({ artistId }, { limit: ARTIST_ALBUMS_LIMIT }));
```
rosscreighton: why not show all of the albums for an artist?

imports/redux/ducks/artist.js
```diff
@@ -1,14 +1,59 @@
...
+
+  return combinedAlbums;
+};
+
+export const setArtistAlbums = ({
+  albums,
+  spotifyId,
+  exclude,
+}) => dispatch => {
+  if (albums.length < ARTIST_ALBUMS_LIMIT) {
```
rosscreighton: again i have questions about the 5 album limit

rosscreighton: we should just talk about it on the phone. just curious how you landed on the limit.

imports/redux/ducks/index.js
```diff
@@ -14,4 +15,5 @@ export default combineReducers({
   ComposeReview,
   Modal,
   Error,
```
rosscreighton: is it a problem that we are using a js reserved keyword `Error` as a key in our state?

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -33,17 +36,41 @@ function AlbumPage({
...
-        </FrostedPane>
-        {reviews.map((review, i) => (
-          <FrostedPane key={i}>
-            <Review {...review} />
+        <div className="AlbumPage-side">
+          <SectionHeader>
+            Tracks
+          </SectionHeader>
+          <div className="gm-list">
+            {album.tracks && album.tracks.map((track, i) => (
```
rosscreighton: i would prefer to assign `tracks` to a prop instead of referencing a key on the album object. i feel like we don't want to depend on the `album` "duck type" having the key `tracks` in this component.

rosscreighton: also, if we pass in `tracks` as a prop, we can use a defaultProp to avoid doing the && check in the code.

imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -68,27 +95,42 @@ function AlbumPageLoader({
...
 
     if (albumHandle.ready()) {
       const album = Albums.findOne({ slug });
-      const reviews = Reviews.find({ albumId: album._id }, { sort: { createdAt: -1 } }).fetch();
-      dispatch(setReviews(reviews));
+
+      dispatch(setAlbum(album));
+      dispatch(requestSpotifyAlbum(album.spotifyId));
+
+      Tracker.autorun(() => {
```
rosscreighton: why did you stop using the subscription callbacks?

rosscreighton: why do all of these things need to be wrapped in autorun blocks?

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: same thing for line 52

imports/ui/components/Track/Track.jsx
```diff
@@ -0,0 +1,45 @@
...
+  name: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  previewUrl: PropTypes.string.isRequired,
+  previewTrackId: PropTypes.func.isRequired,
+  setPreviewTrackId: PropTypes.string.isRequired,
+};
+
+export default connect(state => ({
+  previewTrackId: state.Track.previewTrackId,
+}), dispatch => ({
```
rosscreighton: do you prefer not to use the shorthand for the mapDispatchToProps argument to connect?

imports/ui/components/Track/Track.jsx
```diff
@@ -0,0 +1,45 @@
+import React, { PropTypes } from 'react';
+import { connect } from 'react-redux';
+import { setPreviewTrackId } from '../../../redux/ducks/Track';
+import { rightBlackPointer } from '../../../lib/symbolCodes';
+
+export function Track({
+  name,
+  id,
+  previewUrl,
+  previewTrackId,
```
rosscreighton: what does previewTrackId do?

imports/ui/components/Track/Track.jsx
```diff
@@ -0,0 +1,45 @@
...
+        <div className="gm-list-item-content">
+          <span>{name}</span>
+          <button
+            className="Track-preview"
+            onClick={() => setPreviewTrackId(id)}>{rightBlackPointer}
+          </button>
+        </div>
+      </div>
+      {previewTrackId === id &&
+        <audio className="Track-audio-player" controls>
```
rosscreighton: fuck yeah dude i've never even seen this html element before. i'll be interested to see how it works out for us.

imports/ui/stylesheets/global.scss
```diff
@@ -29,3 +29,33 @@ textarea {
   background-color: $color-smokey-topaz;
   color: $color-white;
 }
+
+// to be used with non list tags
+.gm-list {
+  counter-reset: item;
```
rosscreighton: never heard of this css property. this seems pretty dope 👍

imports/ui/stylesheets/mixins.scss
```diff
@@ -13,3 +13,19 @@
     lost-center: 90% 0px;
   }
 }
+
+@mixin one-fourth() {
```
rosscreighton: do you think we really need a mixin layer on top of lost grid? it seems like maybe an over abstraction to me.

imports/ui/stylesheets/mixins.scss
```diff
@@ -13,3 +13,19 @@
...
+  lost-column: 1/4;
+}
+
+@mixin three-fourths() {
+  lost-column: 3/4;
+}
+
+@mixin gm-list-item-borders() {
```
rosscreighton: may be confused how this mixin is intended to be used, but can't we just add these styles to the gm-list-item class in global.scss?

imports/ui/components/Track/Track.jsx
rosscreighton: looks great in chrome but i wonder how this element looks in other browsers.

imports/ui/stylesheets/components/AlbumPage.scss
```diff
@@ -11,3 +11,13 @@
...
+
+.AlbumPage-side {
+  @include one-fourth();
+  display: flex;
+  flex-direction: column;
+}
+
+.AlbumPage-main {
+  @include three-fourths();
+}
```
rosscreighton: we need to hide the side panel on tablet and below

imports/redux/ducks/Track.js
rosscreighton: ok i get it, i was confused what previewTrackId does. i think `previewTrackId` probably belongs on `AlbumPage`'s state.

imports/ui/components/Track/Track.jsx
rosscreighton: nvm

imports/redux/ducks/Track.js
sameckmeier: I jumped between where it should live and landed on Track for the only reason that it feels like it exposes unnecessary context to AlbumPage. Could be off on my thinking about this, but it feels like previewTrackId is contextual to Track but not vital to AlbumPage. I'd like to avoid adding to page components when possible. Happy to change this if you feel strongly. Either way, it'd probs be good to chat about to which states we connect AlbumPage.

imports/lib/symbolCodes.js
sameckmeier: im pretty sure there's a white one too

imports/api/Albums/server/publications.js
sameckmeier: so this is a holdover for a better solution on the albums we want to show. there's a lot random albums that spotify returns that are associated to an artist like mixes etc we're currently sorting by popularity.

i could imagine having a tab that toggled a different view where reviews are currently that display an artists albums like spotify does (checker board layout) we could show more with the current layout, but didnt want to spend too much time on it as i figured we might want to change this.

imports/redux/ducks/index.js
sameckmeier: in this case i dont believe so (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_keywords_as_of_ECMAScript_6) but im down to change just to be certain.

sameckmeier: good catch 👍

imports/ui/components/AlbumPage/AlbumPage.jsx
sameckmeier: lol my ruby is showing. how about this as a rule of thumb: an object can have an array to keys, so long as it isnt one that contains objs. [1,2,3,4] 👍  [{},{},{}] 👎 

what do ya think?

imports/ui/components/AlbumPage/AlbumPage.jsx
sameckmeier: they dont need to be but they give us finer control over their reactivity. if an embedded autorun inside of an autorun is triggered, it will only run that embedded autorun. if the top level autorun is triggered it triggers all embedded ones. They all get cleaned up the same ie subscriptions are closed in the top level and embedded autoruns.

imports/ui/components/Track/Track.jsx
sameckmeier: i havent tested what it looks like, but it has pretty good support http://www.w3schools.com/tags/tag_audio.asp

imports/ui/components/Track/Track.jsx
sameckmeier: as in do i prefer it over shorthand?

imports/ui/stylesheets/mixins.scss
sameckmeier: no, but added this in case we change out our grid system. i removed this in a later commit because i ended up using flex for AlbumPages layout -- lost doesnt have a succinct way to mix a fixed column with responsive columns.

imports/ui/stylesheets/mixins.scss
sameckmeier: ya changed this one too for the same reasons 👍 misunderstood something about :last-of-type

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: +1 for that rule of thumb 👍

imports/redux/ducks/ViewportChecker.js
```diff
@@ -0,0 +1,25 @@
+export const SET_IS_MOBILE_VIEWPORT = '/gm/Page/set-is-mobile-viewport';
```
rosscreighton: want to talk about ViewportChecker on the phone

rosscreighton:
Per our phone convo about this branch we want to: 

1. change `Track` state to `Tracks` state.
2. remove album display limit entirely, implement "show more" button later.
3. remove ViewportChecker entirely. Use css to show/hide side panel.
4. don't remove Lost grid. remove home-rolled grid classes. use media queries to apply different lost-column widths to the side panel and main section at different screen sizes.

PR #57, opened 2016-09-14
Login ui

rosscreighton:


sameckmeier:
looks fantastic!

blocking: 
- talking about/making SectionToggler stateless and pushing state into redux
- only shallow rendering Authentication/RegistrationForm for tests

curious about:
- sizing issue for Modal close button

rosscreighton:
addressing your comments:

- Because there will be multiple SectionToggle components on the page at once, having this components state in redux becomes more complicated. Imagine if every SectionToggle on the page was subscribed to the `{ SectionToggle: { position: 0 } }` state. When when SectionToggle was toggled, all the others would be toggled as well. To make this component work with redux, we would have to assign each SectionToggle a unique ID when we use it on a page. Though it may be a bit more complicated, i think this is probably what we should do. What do you think?
- 
- AuthenticationForm and RegistrationForm will each have more involved tests when the backend is actually hooked up, so we won't have the silly tests that make sure the button is rendered.

sameckmeier:
i believe we're saying the same thing but just want to make sure im on the same page as ya: for SectionToggle reducer would have different states itd update ie SET_MODAL, SET_MOBILE_TAB, etc
and we'd pass one of those actions to it as a prop which it would then dispatch a long with the corresponding position?

PR #58, opened 2016-09-18
Review avg rating count

sameckmeier:


imports/ui/components/AlbumPage/AlbumPage.jsx
```diff
@@ -87,8 +89,12 @@ function AlbumPageLoader({
 
     if (albumHandle.ready()) {
       const album = Albums.findOne({ slug });
-      const reviews = Reviews.find({ albumId: album._id }, { sort: { createdAt: -1 } }).fetch();
-      dispatch(setReviews(reviews));
+      dispatch(setAlbum(album));
+      dispatch(requestSpotifyAlbum(album.spotifyId));
+
+      const reviews = Reviews.find({ albumId: album._id }, { sort: { createdAt: -1 } });
+      dispatch(setReviewCount(reviews.count()));
```
rosscreighton: should we be doing this on the client? if we have lots of reviews, we won't always have all of those review documents on the client at once, so the review count could be incorrect.

imports/api/reviews/Reviews.js
```diff
@@ -12,4 +13,25 @@ export const schema = new SimpleSchema({
...
+    const album = Albums.findOne(doc.albumId);
+    const reviews = Reviews.find({ albumId: doc.albumId });
+    const reviewsCount = reviews.count();
+
+    let averageRating;
+
+    if (reviewsCount > 1) {
+      const oldReviewCount = reviewsCount - 1;
+      const oldRatingAverage = album.averageRating;
+      averageRating = Math.round(((oldRatingAverage * oldReviewCount) + doc.rating) / reviewsCount);
```
rosscreighton: this was clever 👍

imports/ui/components/AlbumPage/AlbumPage.jsx
rosscreighton: the other option would be storing the count on the Album collection

sameckmeier: good catch, brain farted on that not so tiny detail -- good rule of thumb is to calc where you can instead of store but in this case we def want to store. ill change this one.

rosscreighton:
@sameckmeier this has been approved for awhile so i'm going to :shipit:

PR #78, opened 2016-10-03
moves build system to webpack

rosscreighton:


package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "gm",
   "private": true,
+  "main": "server/main.js",
```
sameckmeier: just to be sure, once we remove meteor, we'll remove the server dirs from the app?

rosscreighton: @sameckmeier we're going to remove this repository entirely. i'll copy over any react/client specific code to the new repo and the the directory structure will probably look completely different.

